PR #7, opened 2021-12-27
OSOE-57: Fixing Meziantou.Analyzer violations

Piedone:


Lombiq.Hosting.Tenants.Admin.Login/Filters/TenantsIndexFilter.cs
```diff
@@ -42,8 +42,8 @@ public async Task OnResultExecutionAsync(ResultExecutingContext context, ResultE
             var actionRouteArea = context.ActionDescriptor.RouteValues["Area"];
             var actionRouteValue = context.ActionDescriptor.RouteValues["Action"];
 
-            if (actionRouteController == typeof(AdminController).ControllerName() &&
-                actionRouteArea == $"{nameof(OrchardCore)}.{nameof(OrchardCore.Tenants)}" &&
+            if (actionRouteController.EqualsOrdinal(typeof(AdminController).ControllerName()) &&
```
sarahelsaig: Would it be worth adding an `IsController<TController>()` extension method instead?

Piedone: I could swear we have something like this but alas we don't it seems. Anyway, it would help here since we want to check if the current controller is one specific controller, not just generally if a class is a controller.

sarahelsaig: okay then

sarahelsaig (approved):

PR #10, opened 2022-03-10
OSOE-60: .NET 6 and Orchard Core 1.3 upgrade

Piedone:
OSOE-60

### DIFF
--- a/Lombiq.Hosting.Tenants.Admin.Login/Services/TenantLoginKeyValidator.cs
+++ b/Lombiq.Hosting.Tenants.Admin.Login/Services/TenantLoginKeyValidator.cs
@@ -19,8 +19,9 @@ namespace Lombiq.Hosting.Tenants.Admin.Login.Services
             using var random = RandomNumberGenerator.Create();
             byte[] uintBuffer = new byte[sizeof(uint)];
 
-            while (length-- > 0)
+            while (length > 0)
             {
+                length--;
                 random.GetBytes(uintBuffer);
                 uint randomNumber = BitConverter.ToUInt32(uintBuffer, 0);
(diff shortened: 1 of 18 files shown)


Lombiq.Hosting.Tenants.Admin.Login/Services/TenantLoginKeyValidator.cs
```diff
@@ -16,16 +16,14 @@ private static string GeneratePassword(int length)
...
 
-                while (length-- > 0)
-                {
-                    random.GetBytes(uintBuffer);
-                    uint randomNumber = BitConverter.ToUInt32(uintBuffer, 0);
-                    stringBuilder.Append(validCharacters[(int)(randomNumber % (uint)validCharacters.Length)]);
-                }
+            while (length-- > 0)
+            {
+                random.GetBytes(uintBuffer);
```
sarahelsaig: The broken arrow in the `while` makes me feel uncomfortable. Could you move the decrementation into a separate line?
```suggestion
            while (length > 0)
            {
                length--;
                random.GetBytes(uintBuffer);
```

sarahelsaig (changes requested):

Lombiq.Hosting.Tenants.Admin.Login/Services/TenantLoginKeyValidator.cs
Piedone: Since in military lingo broken arrow means "an accident that involves nuclear weapons or nuclear weapons components, but does not create the risk of nuclear war" I think you being uncomfortable is justified.

sarahelsaig (approved):

PR #22, opened 2022-07-29
NEST-302: Idle tenant shutdown

Psichorex:
NEST-302


Lombiq.Hosting.Tenants.IdleTenantManagement/Extensions/RuntimeQuotaManagementShellSettingsExtensions.cs
```diff
@@ -0,0 +1,28 @@
+﻿using OrchardCore.Environment.Shell;
```
wAsnk: Ez mehet valami settings modelbe, ahogy beszéltük.

Lombiq.Hosting.Tenants.IdleTenantManagement/Manifest.cs
```diff
@@ -0,0 +1,17 @@
...
+[assembly: Module(
+    Name = "Lombiq Hosting - Idle Tenant Management",
+    Author = "Lombiq Technologies",
+    Website = "https://github.com/Lombiq/Hosting-Tenants",
+    Version = "0.0.1",
+    Description = "Managing idle tenants."
+)]
+
+[assembly: Feature(
+    Id = DisableIdleTenants,
```
wAsnk, line 13: Ahogy beszéltük ez lehetne alwayson feature. És dependencyben legyen ott a background tasks feature.

wAsnk (changes requested):

Lombiq.Hosting.Tenants.IdleTenantManagement/Services/IdleShutdownTask.cs
```diff
@@ -0,0 +1,73 @@
...
+using OrchardCore.Entities;
+using OrchardCore.Environment.Shell;
+using OrchardCore.Environment.Shell.Models;
+using OrchardCore.Modules;
+using OrchardCore.Settings;
+
+namespace Lombiq.Hosting.Tenants.IdleTenantManagement.Services;
+
+[BackgroundTask(Schedule = "* * * * *", Description = "Disable Idle Tenants.")]
+public class IdleShutdownTask : IBackgroundTask
```
Piedone, line 12: This is shutting down the current tenant, with the max idle minutes settings coming from the current tenant's settings. However, we don't want a tenant to be able to manage its own shutdown configuration (think DotNest).

The bg task can run on the tenant (but then on DotNest it needs to be always enabled) but the configuration should not be possible to edit there.

Lombiq.Hosting.Tenants.IdleTenantManagement/Models/IdleMinutesSettings.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Lombiq.Hosting.Tenants.IdleTenantManagement.Models;
+
+public class IdleMinutesSettings
```
Piedone: This is better as an appSettings config, which is also possible to manage on a per-tenant basis. See `DemoSettings` in the Training Demo (but that also uses site settings, which are not needed here).

Lombiq.Hosting.Tenants.IdleTenantManagement/ViewModels/IdleMinutesSettingsViewModel.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Lombiq.Hosting.Tenants.IdleTenantManagement.ViewModels;
```
wAsnk: Ez a fájl már nincs használatban.

wAsnk (changes requested):

Lombiq.Hosting.Tenants.IdleTenantManagement/Manifest.cs
```diff
@@ -0,0 +1,19 @@
...
+    Name = "Lombiq Hosting - Idle Tenant Management - Disable Idle Tenants",
+    Category = "Hosting",
+    Priority = "9999",
```
wAsnk, line 16: Is this needed?

wAsnk (changes requested):

Lombiq.Hosting.Tenants.IdleTenantManagement/Manifest.cs
Psichorex: Yes this one is enabling the feature.

wAsnk: The priority?

Psichorex: The backgroundTask should be the last to run amongst the bg task because if it kills the shell prior to everything else finish it could cause problems in O1 there was also some settings that set the priority to last.

PR #41, opened 2022-10-28
NEST-358: Finishing FeaturesGuard

porgabi:
NEST-358


Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -30,53 +38,193 @@ public sealed class FeaturesEventHandler : IFeatureEventHandler
...
+            foreach (var separatedValue in separatedValues)
+            {
+                valueFormatted.Add(separatedValue.Trim());
+            }
+
+            if (valueFormatted.Contains(featureInfo.Id))
+            {
+                conditionalFeatureIds.Add(keyValuePair.Key);
+            }
+        });
```
Piedone: Use a `foreach` instead, LINQ is confusing for such a mutating code.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -30,53 +38,193 @@ public sealed class FeaturesEventHandler : IFeatureEventHandler
...
+
+        var conditionalFeatures = allFeatures.Where(feature => conditionalFeatureIds.Contains(feature.Id));
+        var currentlyDisabledFeatures = await _shellFeaturesManager.GetDisabledFeaturesAsync();
```
Piedone: What happens if you just enable all the features that need to be kept enabled without checking if they need to be enabled first? Isn't Orchard doing that already? Similarly in `DisableConditionallyEnabledFeaturesAsync`.

Lombiq.Hosting.Tenants.FeaturesGuard/Startup.cs
```diff
@@ -1,31 +1,14 @@
...
-    public Startup(IShellConfiguration shellConfiguration) => _shellConfiguration = shellConfiguration;
-
-    public override void ConfigureServices(IServiceCollection services)
-    {
-        services.Configure<AlwaysEnabledFeaturesOptions>(options =>
-            _shellConfiguration
-                .GetSection("Lombiq_Hosting_Tenants_FeaturesGuard:AlwaysEnabledFeaturesOptions")
-                .Bind(options));
-
+    public override void ConfigureServices(IServiceCollection services) =>
```
Piedone: Both `OrchardCoreBuilderExtensions` methods should rather be called here.

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
```diff
@@ -4,37 +4,42 @@
...
+A module that makes it possible to conditionally enable and conditionally keep enabled, or entirely prevent enabling, configurable sets of features on tenants.
 
 ## Documentation
 
-- To use this feature, enable it on both the Default and the user tenant.
-- Once enabled on the user tenant, the Features Guard feature cannot be disabled.
-- Features that should not be deactivatable can be specified in the appsettings JSON file using `AlwaysEnabledFeaturesOptions`. Example configuration:
+### Conditionally enabling features
+
+- To use this feature, enable it on both the Default and the user tenant, and make sure the app is hosted on Azure.
```
Piedone: Azure is not needed for the module.

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
```diff
@@ -4,37 +4,42 @@
...
 ```
 
-- Additionally, the feature ensures whenever the OrchardCore.Media feature is enabled, the following features also get enabled:
-  - OrchardCore.Content.Types
-  - OrchardCore.Liquid
-  - OrchardCore.Media.Azure.Storage
-  - OrchardCore.Media.Cache
-  - OrchardCore.Settings
+Example case 1:
+- Enables `Azure Storage` when `Media` is enabled. Keeps `Azure Storage` enabled as long as `Media` is enabled, with the exception where if one of `Azure Storage`'s dependencies is disabled, `Azure Media` does not get re-enabled.
```
Piedone: In this example, Azure Storage shouldn't be disabled in any case, including if any of its dependencies get disabled, as long as Media is enabled. Otherwise, people can easily defeat this.

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
```diff
@@ -4,37 +4,42 @@
...
+
+Example case 2:
+- Enables `Twitter` when `Media` or `Workflows` is enabled. Keeps `Twitter` enabled as long as either `Media` or `Workflows` remains enabled, with the exception where if one of `Twitter`'s dependencies is disabled, `Twitter` does not get re-enabled.
+
+### Preventing enabling features
 
-- Preventing enabling certain features on user tenants is also possible via recipes in a FeatureProfiles step. Example configuration:
+- Preventing enabling certain features on user tenants is possible via recipes in a `FeatureProfiles` step. Example configuration:
```
Piedone: Link to the official docs.

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
```diff
@@ -4,37 +4,42 @@
...
-  - OrchardCore.Media.Azure.Storage
-  - OrchardCore.Media.Cache
-  - OrchardCore.Settings
+Example case 1:
+- Enables `Azure Storage` when `Media` is enabled. Keeps `Azure Storage` enabled as long as `Media` is enabled, with the exception where if one of `Azure Storage`'s dependencies is disabled, `Azure Media` does not get re-enabled.
+
+Example case 2:
+- Enables `Twitter` when `Media` or `Workflows` is enabled. Keeps `Twitter` enabled as long as either `Media` or `Workflows` remains enabled, with the exception where if one of `Twitter`'s dependencies is disabled, `Twitter` does not get re-enabled.
+
+### Preventing enabling features
```
Piedone, line 56: The "Contributing and support" section should only be there in the root Readme.

Piedone (changes requested):

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
porgabi: Artist's depiction of users defeeting FeaturesGuard:
image

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
porgabi: Well, the universe has not yet imploded despite my best efforts while testing it without these checks. It looks like they can be removed indeed.

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
Piedone: :D * insert comic of soldier defending sleeping person, but badly *

porgabi: image

Lombiq.Hosting.Tenants.FeaturesGuard/Startup.cs
porgabi: Akin to _IdleTenantManagement_'s config, without `OrchardCoreBuilder`?
image


Or is there a way to use the `OrchardCoreBuilderExtension` methods inside a _Startup_ class too? `services.AddOrchardCore()` seems to result in an exception.
image

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
Piedone: Thank you :D.

Lombiq.Hosting.Tenants.FeaturesGuard/Startup.cs
Piedone: How it is for `IdleTenantManagement`, yes.

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
0liver: I think, you should pursue a comic writer career - at least on the side, @porgabi.

Lombiq.Hosting.Tenants.FeaturesGuard/Startup.cs
porgabi: The problem I'm encountering now is during UI test runs, `_shellConfiguration.IsUITesting();` returns `false` in FeaturesGuard's _Startup_, while the same method returns true in _Program_.

EDIT -- seems to be different between _IShellConfiguration_ and _IConfiguration:_
image

Guess _IConfiguration_ it is.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -93,35 +93,27 @@ public async Task EnableConditionallyEnabledFeaturesAsync(IFeatureInfo featureIn
...
-        var currentlyDisabledConditionalFeatures = currentlyDisabledFeatures.Intersect(conditionalFeatures);
-        if (currentlyDisabledConditionalFeatures.Any())
+        // During setup, Shell Descriptor can become out of sync with the DB when it comes to enabled features,
+        // but it's more accurate than IShellDescriptorManager's methods.
+        var shellDescriptor = await _shellDescriptorManager.GetShellDescriptorAsync();
+
+        // If Shell Descriptor's Features already contains a feature that is found in conditionalFeatures, remove it
+        // from the list.
+        var featuresToEnable = conditionalFeatures.ToList();
+        conditionalFeatures.ForEach(feature =>
```
Piedone: Same issue with this foreach too.

Lombiq.Hosting.Tenants.FeaturesGuard/Startup.cs
```diff
@@ -1,14 +1,44 @@
...
+                    .Bind(options));
+        }
+
+        if (_configuration.IsUITesting())
+        {
+            services.Configure<ConditionallyEnabledFeaturesOptions>(options =>
+                _shellConfiguration
+                    .GetSection("Lombiq_Hosting_Tenants_FeaturesGuard:TestConditionallyEnabledFeaturesOptions:ConditionallyEnabledFeatures")
+                    .Bind(options));
+        }
```
Piedone: These again tie the module to Azure (keep in mind that conditionally enabled features is not something just interesting for Azure), and the UI Testing Toolbox as well. Neither of these is needed.

Instead, these `Configure()` calls (or rather, only the first one, no need for a separate testing) can be here without conditions. And in the `Program`/`Startup` of the web app, you can add the configuration conditionally, as opposed to it being statically in appsettings. E.g.:

```c#
builder.Services
    .AddSingleton(configuration)
    .AddOrchardCms(orchardCoreBuilder =>
    {
        // ...

        if (configuration.IsAzureHosting())
        {
            orchardCoreBuilder.ConfigureServices(services =>
                services.Configure<ConditionallyEnabledFeaturesOptions>(options =>
                    options.EnableFeatureIfOtherFeatureIsEnabled = new Dictionary<string, string>
                    {
                        ["OrchardCore.Media.Azure.Storage"] = "OrchardCore.Media",
                    }));
        }
    });
```

Since we'll repeatedly use this, you can also add an extension method similar to `AllowMiniProfilerOnAdmin()` in Orchard, so one doesn't need to write the whole ceremony, just a) set `EnableFeatureIfOtherFeatureIsEnabled` if you have something custom or b) set the Azure Storage config directly since that'll be a common use-case. (I realize that this brings something from Azure into the module, but it doesn't require it to statically reference anything from Azure and still keeps the option open for any other config, like for Elasticsarch).

After this, configs from both sources (code, appsettings) will be merged, so you don't need the latter (but we keep the option open to use it). And you can remove the unneeded project/NuGet references from this project.

Piedone (changes requested):

Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class OrchardCoreBuilderExtensions
+{
+    /// <summary>
+    /// Binds Azure Storage as the conditional feature and Media as the condition feature to
+    /// <see cref="ConditionallyEnabledFeaturesOptions"/>.
+    /// </summary>
+    public static OrchardCoreBuilder ConfigureFeaturesGuardForAzureStorage(this OrchardCoreBuilder builder)
```
Piedone, line 12: Document this as an alternative configuration option in the Readme.

Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
```diff
@@ -0,0 +1,23 @@
+using Lombiq.Hosting.Tenants.FeaturesGuard.Models;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class OrchardCoreBuilderExtensions
```
Piedone, line 6: As I've written [here](https://github.com/Lombiq/Hosting-Tenants/pull/41#discussion_r1012341787):

> Since we'll repeatedly use this, you can also add an extension method similar to AllowMiniProfilerOnAdmin() in Orchard, so one doesn't need to write the whole ceremony, just **a) set EnableFeatureIfOtherFeatureIsEnabled if you have something custom** or b) set the Azure Storage config directly since that'll be a common use-case. (

Lombiq.Hosting.Tenants.FeaturesGuard.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
```diff
@@ -19,19 +19,35 @@ public static async Task TestForbiddenFeaturesAsync(this UITestContext context,
...
+        // When UIKit gets disabled and ChartJs is also disabled, Twitter should get disabled.
+        await context.ClickReliablyOnAsync(By.XPath("//a[@id='btn-disable-Lombiq_UIKit']"));
         await context.ClickModalOkAsync();
-        context.Exists(By.XPath("//a[@id='btn-disable-OrchardCore_Users']"));
+        context.Exists(By.XPath("//a[@id='btn-enable-Lombiq_ChartJs']"));
+        context.Exists(By.XPath("//a[@id='btn-enable-OrchardCore_Twitter']"));
+
+        // When UIKit is enabled, Twitter should get enabled.
+        await context.ClickReliablyOnAsync(By.XPath("//a[@id='btn-enable-Lombiq_UIKit']"));
+        context.Exists(By.XPath("//a[@id='btn-disable-OrchardCore_Twitter']"));
```
Piedone, line 55: Also test that Twitter can't be disabled why UIKit and ChartJs are enabled.

Lombiq.Hosting.Tenants.FeaturesGuard/Models/ConditionallyEnabledFeaturesOptions.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace Lombiq.Hosting.Tenants.FeaturesGuard.Models;
+
+public class ConditionallyEnabledFeaturesOptions
+{
+    // Needs to be settable for binding.
+#pragma warning disable CA2227 // Collection properties should be read only
+    public IDictionary<string, string> EnableFeatureIfOtherFeatureIsEnabled { get; set; }
```
Piedone: The value type would be more applicable with a string array or `IEnumerable<string>` than comma-separated string values.

Piedone (changes requested):

Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
porgabi: Yes, this is what I intended to do with `ConfigureFeaturesGuardForAzureStorage()` -- so it's possible to set the Azure Storage config with this extension method, or add custom configuration in _appsettings_. Which part of this did I get wrong?

Piedone: In addition to this, add a second extension method to set EnableFeatureIfOtherFeatureIsEnabled if you have something custom, not related to Azure and Media.

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
```diff
@@ -11,9 +11,25 @@ A module that makes it possible to conditionally enable and conditionally keep e
 ### Conditionally enabling features
 
 - To use this feature, enable it on both the Default and the user tenant.
-- Features that should be conditionally enabled, as well as the features whose status acts as the condition, can be specified in appsettings.json using `ConditionallyEnabledFeaturesOptions`.
+- Features that should be conditionally enabled, as well as the features whose status acts as the condition, can be specified either in appsettings.json using `ConditionallyEnabledFeaturesOptions` or in Program via the `ConfigureFeaturesGuard()` extension method.
```
Piedone:
```suggestion
- Features that should be conditionally enabled, as well as the features whose status acts as the condition, can be specified either in _appsettings.json_ using `ConditionallyEnabledFeaturesOptions` or in `Program` via the `ConfigureFeaturesGuard()` extension method.
```

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
```diff
@@ -11,9 +11,25 @@ A module that makes it possible to conditionally enable and conditionally keep e
...
 Conditionally enabled features need to be provided with a singular or multiple condition features, where the status of the condition features determines whether the corresponding conditional feature is enabled or disabled. Example configuration:
 
+Program config example:
```
Piedone:
```suggestion
`Program` config example:
```

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
```diff
@@ -11,9 +11,25 @@ A module that makes it possible to conditionally enable and conditionally keep e
...
+            {
+                // "Enable this feature and keep it enabled": "If any of these features are enabled",
+                ["OrchardCore.Media.Azure.Storage"] = new List<string> { "OrchardCore.Media" },
+                ["OrchardCore.Twitter"] = new List<string> { "Lombiq.UIKit", "Lombiq.ChartJs" },
+            }));
+```
+
+- `ConfigureFeaturesGuardForAzureStorage()` is also available to add `Azure Storage` as a conditional feature and `Media` as its condition feature.
+
+appsettings.json config example:
```
Piedone:
```suggestion
_appsettings.json_ config example:
```

Lombiq.Hosting.Tenants.FeaturesGuard/Readme.md
```diff
@@ -11,9 +11,25 @@ A module that makes it possible to conditionally enable and conditionally keep e
...
 
+Program config example:
+```
+builder.Services
+    .AddOrchardCms(orchardCoreBuilder =>
+        orchardCoreBuilder.ConfigureFeaturesGuard(
+            new Dictionary<string, IEnumerable<string>>
+            {
+                // "Enable this feature and keep it enabled": "If any of these features are enabled",
+                ["OrchardCore.Media.Azure.Storage"] = new List<string> { "OrchardCore.Media" },
```
Piedone: Below too.

```suggestion
                ["OrchardCore.Media.Azure.Storage"] = new[] { "OrchardCore.Media" },
```

Lombiq.Hosting.Tenants.FeaturesGuard.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
```diff
@@ -39,6 +39,11 @@ public static async Task TestConditionallyEnabledFeaturesAsync(this UITestContex
         context.Exists(By.XPath("//a[@id='btn-disable-Lombiq_UIKit']"));
         context.Exists(By.XPath("//a[@id='btn-disable-OrchardCore_Twitter']"));
 
+        // When either UIKit or ChartJs is enabled, Twitter should not be deactivatable.
```
Piedone:
```suggestion
        // When either UIKit or ChartJs is enabled, Twitter should not be possible to disable.
```

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -65,10 +65,9 @@ public async Task EnableConditionallyEnabledFeaturesAsync(IFeatureInfo featureIn
         }
 
         var allConditionFeatureIds = new List<string>();
-        foreach (var conditionFeatureId in conditionallyEnabledFeatures.Values)
```
Piedone: It's not a list, nor should you use such Hungarian notation if avoidable. In this case, it can simply be `conditionFeatureIds`.

Piedone (changes requested):

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
porgabi: Don't think it's got anything to do with Hungarian notations, I would think it's the same as saying "shopping list"; it doesn't necessarily have to be "list of something". 

But it's not a list indeed so I'll change the name.

PR #44, opened 2022-12-13
NEST-362: Adding features guard configuration for Elasticsearch

DemeSzabolcs:
[NEST-362](https://lombiq.atlassian.net/browse/NEST-362)

### DIFF
--- a/Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
+++ b/Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
@@ -1,4 +1,5 @@
 using Lombiq.Hosting.Tenants.FeaturesGuard.Models;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using System.Collections.Generic;
 
 namespace Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
```diff
@@ -21,6 +21,23 @@ public static OrchardCoreBuilder ConfigureFeaturesGuardForAzureStorage(this Orch
...
 
+    /// <summary>
+    /// Binds Elasticsearch as the conditional feature and Search or Indexing as the condition feature to
+    /// <see cref="ConditionallyEnabledFeaturesOptions"/>.
+    /// </summary>
+    public static OrchardCoreBuilder ConfigureFeaturesGuardForElasticsearch(this OrchardCoreBuilder builder)
+    {
+        builder.ConfigureServices((tenantServices, _) =>
+            tenantServices.PostConfigure<ConditionallyEnabledFeaturesOptions>(options =>
+                options.EnableFeatureIfOtherFeatureIsEnabled = new Dictionary<string, IEnumerable<string>>
```
wAsnk: Change the logic so if the dictionary EnableFeatureIfOtherFeatureIsEnabled already exists add the `["OrchardCore.Search.Elasticsearch"] =
                        new List<string> { "OrchardCore.Search", "OrchardCore.Indexing" },` key value pair into it, instead of overriding the value to this new dictionary.

Same for `ConfigureFeaturesGuardForAzureStorage()`. Then you can use both functions after each other.

wAsnk (changes requested):

Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
DemeSzabolcs: Good idea!

wAsnk: Also I think you can dry these two together.

DemeSzabolcs: > Also I think you can dry these two together.

I don't want to put them into one configure function, because we have a separate `if` for adding the Azure configuration: 
image

So not always both are enabled.

wAsnk: Yeah I know. But you can DRY it without using both.

wAsnk:
Please update the readme where this part is mentioned.

wAsnk (approved):

PR #46, opened 2023-01-02
NEST-390: Elastic Search is not enabled when required

DemeSzabolcs:
[NEST-390](https://lombiq.atlassian.net/browse/NEST-390)

### DIFF
--- a/Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
+++ b/Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
@@ -1,5 +1,6 @@
 using Lombiq.Hosting.Tenants.FeaturesGuard.Models;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.Extensions.DependencyInjection;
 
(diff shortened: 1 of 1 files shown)


Lombiq.Hosting.Tenants.FeaturesGuard/Extensions/OrchardCoreBuilderExtensions.cs
```diff
@@ -55,8 +56,9 @@ public static class OrchardCoreBuilderExtensions
                 }
                 else
                 {
-                    options.EnableFeatureIfOtherFeatureIsEnabled
-                        .AddRange(configDictionary);
+                    options.EnableFeatureIfOtherFeatureIsEnabled.AddRange(
```
wAsnk, line 59: But why does it try to add it twice? Is that known?

DemeSzabolcs: It was this way always. And previously it replaced the whole dictionary, so it was not a problem. See this part from the old code: https://github.com/Lombiq/Hosting-Tenants/pull/44/files#diff-4505e2e637d14d2db863c6aa96560652c30b84897bc1bbdae3ae69c6708de15aL12-L22

I think when you turn on/off a feature, it will reload the tenant and `Program.cs` will run again. It sees that `Media` is enabled, so it tries to enable Azure.Media storage too even if it was enabled. As I mentioned previously the dictionary was replaced, but now we are adding elements to it (`ConfigureFeaturesGuardWithoutOverriding`), so we can have multiple things in it, without recreating it with all the elements.

So in a nutshell it's okay, I just missed this fact in the https://github.com/Lombiq/Hosting-Tenants/pull/44 PR.

wAsnk (approved):

PR #47, opened 2023-01-11
OSOE-532: Fixing analyzer violations

Piedone:
[OSOE-532](https://lombiq.atlassian.net/browse/OSOE-532)

[OSOE-532]: https://lombiq.atlassian.net/browse/OSOE-532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
+++ b/Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
@@ -83,23 +83,14 @@ public sealed class FeaturesEventHandler : IFeatureEventHandler
             .Select(keyValuePair => keyValuePair.Key)
             .ToList();
 
-        // Handle multiple conditional features as well.
-        var conditionalFeatures = allFeatures.Where(feature => conditionalFeatureIds.Contains(feature.Id));
-
         // During setup, Shell Descriptor can become out of sync with the DB when it comes to enabled features,
         // but it's more accurate than IShellDescriptorManager's methods.
         var shellDescriptor = await _shellDescriptorManager.GetShellDescriptorAsync();
 
         // If Shell Descriptor's Features already contains a feature that is found in conditionalFeatures, remove it
-        // from the list.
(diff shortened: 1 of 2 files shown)


Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -92,11 +92,14 @@ public async Task EnableConditionallyEnabledFeaturesAsync(IFeatureInfo featureIn
...
         // If Shell Descriptor's Features already contains a feature that is found in conditionalFeatures, remove it
         // from the list.
+        // Double enumeration is necessary, since we can't remove from the List without breaking enumeration.
+#pragma warning disable CA1851 // Possible multiple enumerations of 'IEnumerable' collection
         var featuresToEnable = conditionalFeatures.ToList();
         foreach (var feature in conditionalFeatures.Where(feature => shellDescriptor.Features.Contains(new ShellFeature(feature.Id))))
         {
             featuresToEnable.Remove(feature);
         }
+#pragma warning restore CA1851 // Possible multiple enumerations of 'IEnumerable' collection
```
sarahelsaig: You actually have multiple enumeration and I don't think it's really necessary, so the warning shouldn't be disabled.

It's not clear to me why `var conditionalFeatures = ...` has to be before `var shellDescriptor = ...`, is that really necessary? I don't see why. If not, then you could switch their places, and then `conditionalFeatures` and `featuresToEnable` can be merged into a single expression, as in `allFeatures.Where(feature => ... && !...).ToList()`.

sarahelsaig (changes requested):

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Piedone: Indeed!

PR #50, opened 2023-02-23
NEST-361: Fix memory leak in idle tenant shutdown

Psichorex:
[NEST-361](https://lombiq.atlassian.net/browse/NEST-361)

[NEST-361]: https://lombiq.atlassian.net/browse/NEST-361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Hosting.Tenants.IdleTenantManagement.Tests.UI/Extensions/IdleTenantManagementExtensions.cs
```diff
@@ -26,5 +26,5 @@ public static class IdleTenantManagementExtensions
     public static readonly Func<IWebApplicationInstance, Task> AssertAppLogsWithIdleCheckAsync =
         async webApplicationInstance =>
             (await webApplicationInstance.GetLogOutputAsync())
-            .ShouldContain("Shutting down tenant \"Default\" because of idle timeout.");
+            .ShouldContain("Shutting down tenant \"IdleTenantShutdown\" because of idle timeout.");
```
wAsnk: Shouldn't the tenant name be an input parameter also like it is in the TestIdleTenantManagerBehaviorAsync? Otherwise, this might fail.

wAsnk (changes requested):

Lombiq.Hosting.Tenants.IdleTenantManagement.Tests.UI/Extensions/IdleTenantManagementExtensions.cs
Psichorex: Imo not really because in the tenant setup of the UITestExtension method this exact name is give for the tenant to be created. So in theory it can't fail. This is just a static way of testing a tenant is indeed shutdown or not. I don't think we need flexible testing.

Lombiq.Hosting.Tenants.IdleTenantManagement.Tests.UI/Extensions/IdleTenantManagementExtensions.cs
```diff
@@ -26,5 +27,5 @@ public static class IdleTenantManagementExtensions
     public static readonly Func<IWebApplicationInstance, Task> AssertAppLogsWithIdleCheckAsync =
         async webApplicationInstance =>
             (await webApplicationInstance.GetLogOutputAsync())
-            .ShouldContain("Shutting down tenant \"Default\" because of idle timeout.");
+            .ShouldContain($"Shutting down tenant \"{IdleTenantShutdown}\" because of idle timeout.");
```
wAsnk: What I've meant here is that the function should have a parameter that is used for the tenant name, just like in TestIdleTenantManagerBehaviorAsync. Because in TestIdleTenantManagerBehaviorAsync it's dynamic and here it is constant.

wAsnk (changes requested):

Lombiq.Hosting.Tenants.IdleTenantManagement.Tests.UI/Extensions/IdleTenantManagementExtensions.cs
wAsnk: Yeah, but then why do you have the option for the test itself to be anything?

wAsnk (approved):

PR #54, opened 2023-04-23
NEST-326: Maintenance tasks for tenants

barthamark:
[NEST-326](https://lombiq.atlassian.net/browse/NEST-326)

[NEST-326]: https://lombiq.atlassian.net/browse/NEST-326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Hosting.Tenants.Maintenance/Readme.md
```diff
@@ -0,0 +1,49 @@
+# Lombiq Hosting - Tenant Maintenance for Orchard Core
+
+## About
+
+With the help of this module you can execute maintenance tasks on tenants.
```
Piedone: Explain what a maintenance task is.

Lombiq.Hosting.Tenants.Maintenance/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+Please see the below features for more information.
+
+### `Lombiq.Hosting.Tenants.Maintenance`
+
+This is the core functionality required to execute maintenance tasks on tenants. It is available on any tenant. To make your application execute maintenance tasks, you need to add the following to your `Startup.cs`:
+
+```csharp
+public void ConfigureServices(IServiceCollection services) =>
+    services.AddOrchardCms(
+        builder => builder.AddTenantFeatures(Lombiq.Hosting.Tenants.Maintenance.Constants.FeatureNames.Maintenance));
```
Piedone, line 20: Is this really needed? Enabled on all tenants? Same below.

Lombiq.Hosting.Tenants.Maintenance/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+It's a maintenance task that updates the tenants' URL based on the app configuration. It is available on any tenant. To make your application execute this task, you need to add the following to your `Startup.cs`:
+
+```csharp
+public void ConfigureServices(IServiceCollection services) =>
+    services.AddOrchardCms(
+        builder => builder.AddTenantFeatures(Lombiq.Hosting.Tenants.Maintenance.Constants.FeatureNames.UpdateTenantUrl));
+```
+
+To configure the task, you need to add the following to your `appsettings.json`:
```
Piedone:
```suggestion
To configure the task, you need to add the following to your _appsettings.json_:
```

Lombiq.Hosting.Tenants.Maintenance/Readme.md
barthamark: Well, you can disable maintenances with per-tenant configuration if the provider supports configuration. If not, then we can say that it's an important maintenance to be executed on all tenants. I'm curious what your thoughts are. In our cases the maintenances should be active on all tenants if applicable (e.g., a certain feature is active that needs to be fixed, or a URL replace). Do you think there's any scenario where it's not needed to be enabled on all tenants?

Piedone: We discussed that the URL replace should rather be something that only the Default tenant does. For something like user configuration, yes, but not necessarily all maintenance should be enabled on all tenants.

barthamark: Yes but that clearly must run only on the default shell which is a somewhat different scenario from having the option to run it on Tenant A, E, X, etc. as it always must be executed on the default shell; hence the .IsDefaultShell() condition in the ShouldExecuteAsync.

Piedone: Why just not enable that particular feature on all tenants? We don't want to run a NOOP on every tenant except the Default one.

barthamark: OK, I agree. However, now I have a related question.

Currently, I have two providers, one iterate through the shell settings and updates the tenant URL - it runs only on the default tenant. The other updates the site settings, but it is executed per tenant. Now if I want to do it on the default tenant, too, I'd need to iterate through each tenant, create a shell context, and execute a DB query with the site settings update which has multiple problems:

1. Performance issue.
2. We address tenants that might be totally inactive.
3. I met an annoying bug while using CreateScope for IShellHost during tenant activation, it causes an infinite loop during UI testing (it caused my literal headache in the past 1.5 weeks, and finally dropped using IShellHost entirely).
4. You don't even have a base URL set in the site settings for any of the representative tenants, and it's not even that important to set in my opinion.

So my suggestion is to either keep the site settings updater per tenant and run the tenant binding replacer on the default tenant; it means that those will be separate features. Or let's remove the site settings base URL updater because it's not that important.

What do you think?

Piedone: >So my suggestion is to either keep the site settings updater per tenant and run the tenant binding replacer on the default tenant; it means that those will be separate features.

Yep!

Lombiq.Hosting.Tenants.Maintenance/License.md
```diff
@@ -0,0 +1,13 @@
+Copyright © 2022, [Lombiq Technologies Ltd.](https://lombiq.com)
```
wAsnk:
```suggestion
Copyright © 2023, [Lombiq Technologies Ltd.](https://lombiq.com)
```

Lombiq.Hosting.Tenants.Maintenance/Lombiq.Hosting.Tenants.Maintenance.csproj
```diff
@@ -0,0 +1,44 @@
...
+    <PropertyGroup>
+        <TargetFramework>net6.0</TargetFramework>
+        <AddRazorSupportForMvc>true</AddRazorSupportForMvc>
+        <DefaultItemExcludes>$(DefaultItemExcludes);.git*;node_modules\**;Tests\**</DefaultItemExcludes>
+    </PropertyGroup>
+
+    <PropertyGroup>
+        <Title>Lombiq Hosting - Tenants Maintenance for Orchard Core</Title>
+        <Authors>Lombiq Technologies</Authors>
+        <Copyright>Copyright © 2022, Lombiq Technologies Ltd.</Copyright>
```
wAsnk:
```suggestion
        <Copyright>Copyright © 2023, Lombiq Technologies Ltd.</Copyright>
```

Lombiq.Hosting.Tenants.Maintenance/Maintenance/TenantUrlMaintenanceCore/Startup.cs
```diff
@@ -0,0 +1,24 @@
...
+    private readonly IShellConfiguration _shellConfiguration;
+
+    public Startup(IShellConfiguration shellConfiguration) =>
+        _shellConfiguration = shellConfiguration;
+
+    public override void ConfigureServices(IServiceCollection services)
+    {
+        var options = new TenantUrlMaintenanceOptions();
+        var configSection = _shellConfiguration.GetSection("Lombiq_Hosting_Tenants_Maintenance:TenantUrlMaintenance");
+        configSection.Bind(options);
```
wAsnk: Isn't it possible to do this in one line?

Lombiq.Hosting.Tenants.Maintenance/Maintenance/TenantUrlMaintenanceCore/TenantUrlHelper.cs
```diff
@@ -0,0 +1,22 @@
+using OrchardCore.Environment.Shell;
+
+namespace Lombiq.Hosting.Tenants.Maintenance.Maintenance.TenantUrlMaintenanceCore;
+
+internal static class TenantUrlHelper
+{
+    public static string GetTenantUrl(TenantUrlMaintenanceOptions options, ShellSettings shellSettings)
+    {
+#pragma warning disable CA1308 // Normalize strings to uppercase
```
wAsnk: Also comment on why it is necessary, to do it this way.

Lombiq.Hosting.Tenants.Maintenance/Services/IMaintenanceProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Gets the order of the maintenance task. The lower the number the earlier the task will be executed.
+    /// </summary>
+    int Order { get; }
+
+    /// <summary>
+    /// Determines whether the maintenance task should be executed.
+    /// </summary>
+    /// <param name="context">Provides information about the current execution.</param>
+    /// <returns>True if the maintenance task should be executed, false otherwise.</returns>
```
wAsnk:
```suggestion
    /// <returns><see langword="true"/> if the maintenance task should be executed, <see langword="false"/> otherwise.</returns>
```

Lombiq.Hosting.Tenants.Maintenance/Services/MaintenanceManager.cs
```diff
@@ -0,0 +1,113 @@
...
+        foreach (var provider in orderedProviders)
+        {
+            var currentExecution = new MaintenanceTaskExecutionData
+            {
+                MaintenanceId = provider.Id,
+                ExecutionTimeUtc = _clock.UtcNow,
+            };
+            var context = new MaintenanceTaskExecutionContext
+            {
+                // GetLatestExecutionTask = () => GetLatestExecutionByMaintenanceIdAsync(provider.Id),
```
wAsnk: Delete it if it's not needed.

Lombiq.Hosting.Tenants.Maintenance/Services/MaintenanceManager.cs
```diff
@@ -0,0 +1,113 @@
...
+
+    public Task<MaintenanceTaskExecutionData> GetLatestExecutionByMaintenanceIdAsync(string maintenanceId) =>
+        _session.Query<MaintenanceTaskExecutionData, MaintenanceTaskExecutionIndex>(collection: DocumentCollections.Maintenance)
+            .Where(execution => execution.MaintenanceId == maintenanceId)
+            .OrderByDescending(execution => execution.ExecutionTimeUtc)
+            .FirstOrDefaultAsync();
+
+    public async Task ExecuteMaintenanceTasksAsync()
+    {
+        var orderedProviders = _maintenanceProviders.OrderBy(p => p.Order);
```
wAsnk: Please use a descriptive lambda variable name.

Lombiq.Hosting.Tenants.Maintenance/Services/MaintenanceManager.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                await provider.ExecuteAsync(context);
+                execution.IsSuccess = string.IsNullOrEmpty(execution.Error);
+                if (execution.IsSuccess)
+                {
+                    _logger.LogDebug("Maintenance task {MaintenanceId} executed successfully.", provider.Id);
+                }
+                else
+                {
+                    _logger.LogError(
```
wAsnk, line 85: I'm not sure, does it log which tenant had the error on?

Lombiq.Hosting.Tenants.Maintenance/Services/MaintenanceRunnerService.cs
```diff
@@ -0,0 +1,35 @@
...
+        _logger = logger;
+        _maintenanceManager = maintenanceManager;
+    }
+
+    public override async Task ActivatedAsync()
+    {
+        if (_shellSettings.State != TenantState.Running) return;
+
+        // Getting the scope here is important because the shell might not be fully initialized yet.
+        _logger.LogInformation(
```
wAsnk: Why do you log on information level here if you log on debug level in the maintenance manager?

Lombiq.Hosting.Tenants.Maintenance/Services/MaintenanceRunnerService.cs
```diff
@@ -0,0 +1,35 @@
...
+        _shellSettings = shellSettings;
+        _logger = logger;
+        _maintenanceManager = maintenanceManager;
+    }
+
+    public override async Task ActivatedAsync()
+    {
+        if (_shellSettings.State != TenantState.Running) return;
+
+        // Getting the scope here is important because the shell might not be fully initialized yet.
```
wAsnk: I think you left this comment here by accident.

Lombiq.Hosting.Tenants.Maintenance/Startup.cs
```diff
@@ -0,0 +1,23 @@
...
+using OrchardCore.Modules;
+using YesSql.Indexes;
+
+namespace Lombiq.Hosting.Tenants.Maintenance;
+
+public class Startup : StartupBase
+{
+    public override void ConfigureServices(IServiceCollection services)
+    {
+        services.Configure<StoreCollectionOptions>(o => o.Collections.Add(DocumentCollections.Maintenance));
```
wAsnk: Use descriptive lambda variable name.

wAsnk (changes requested):

Lombiq.Hosting.Tenants.Maintenance/Readme.md
```diff
@@ -0,0 +1,54 @@
+# Lombiq Hosting - Tenant Maintenance for Orchard Core
+
```
wAsnk, line 2: The NuGet shield is missing here and in the main readme. Also include this feature in the main readme.

wAsnk (changes requested):

Lombiq.Hosting.Tenants.Maintenance/License.md
0liver: I don't think we should adjust those numbers, see https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/blob/dev/License.md, which has 2020 in it.

Piedone: All sub-projects should use the same in the given project, in this case this one: https://github.com/Lombiq/Hosting-Tenants/blob/dev/Lombiq.Hosting.Tenants.Admin.Login/License.md

Lombiq.Hosting.Tenants.Maintenance/Maintenance/TenantUrlMaintenanceCore/Startup.cs
barthamark: No.

Lombiq.Hosting.Tenants.Maintenance/Services/MaintenanceManager.cs
barthamark: Yes.

Lombiq.Hosting.Tenants.Maintenance/License.md
wAsnk: Shouldn't it present the year when the project itself was created?

Lombiq.Hosting.Tenants.Maintenance/Maintenance/TenantUrlMaintenanceCore/Startup.cs
wAsnk: Why `_shellConfiguration.GetSection("Lombiq_Hosting_Tenants_Maintenance:TenantUrlMaintenance").Bind(options);` is not possible?

0liver: Because `configSection` is used in the line below, too.

barthamark: That's correct. 👍

Lombiq.Hosting.Tenants.Maintenance/License.md
Piedone: The project that the license is for here is the whole repository. Any new csprojs to it are extensions of this original project, and thus should use the original license.

wAsnk: Okay, thank you!

Lombiq.Hosting.Tenants.Maintenance/Maintenance/TenantUrlMaintenanceCore/TenantUrlHelper.cs
wAsnk: Please comment the _reason_ not that it is needed because it is necessary 😄

wAsnk:
Please also set the _Url Prefix_ which is required in these cases since every tenant will have the same _hostname_, tenants-staging.dotnest.net in this case.

Lombiq.Hosting.Tenants.Maintenance/Readme.md
```diff
@@ -43,14 +43,50 @@ Provides the core functionality for updating the tenants' URL based on the app c
...
+It's a maintenance task that updates the shell's request URLs in each tenant's shell settings based on the app configuration. It is available only for the default tenant.
+
+The following configuration options are available to set the shell request URLs:
+
+```json
+{
+  "OrchardCore": {
+    "Lombiq_Hosting_Tenants_Maintenance": {
+      "UpdateShellRequestUrl": {
+        "DefaultRequestUrl": "domain.com",
```
wAsnk: I think this should be DefaultShellRequestUrl as per the UpdateShellRequestUrlMaintenanceOptions.

Lombiq.Hosting.Tenants.Maintenance/Readme.md
```diff
@@ -43,14 +43,50 @@ Provides the core functionality for updating the tenants' URL based on the app c
...
 ```
 
+The following configuration options are available to set the site URL:
+
+```json
+{
+  "OrchardCore": {
+    "Lombiq_Hosting_Tenants_Maintenance": {
+      "UpdateSiteUrl": {
+        "SiteUrl": "https://domain.com/{TenantName}"
```
wAsnk: This maintenance would also require a default tenant url, because in our example on dotnest staging it has a different structure than this.

wAsnk (changes requested):

wAsnk (approved):
Looks good to me. Anything you would like to add @Piedone ?

Piedone:
Some testing would be necessary too.

wAsnk (approved):

PR #55, opened 2023-04-25
OSOE-548: Upgrade to Orchard Core 1.6 in Hosting-Tenants

Psichorex:
[OSOE-548](https://lombiq.atlassian.net/browse/OSOE-548)

[OSOE-548]: https://lombiq.atlassian.net/browse/OSOE-548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Piedone (approved):

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -90,9 +88,20 @@ public async Task EnableConditionallyEnabledFeaturesAsync(IFeatureInfo featureIn
...
-            conditionalFeatureIds.Contains(feature.Id) && !shellDescriptor.Features.Contains(new ShellFeature(feature.Id)));
+            conditionalFeatureIds.Contains(feature.Id) &&
+            !shellDescriptor.Features.Any(shellFeature => conditionalFeatureIds.Contains(shellFeature.Id)));
+
+        foreach (var feature in featuresToEnable)
+        {
+            if (_previouslyAddedFeatures.Contains(feature))
+            {
+                return;
+            }
```
Piedone: What if `featuresToEnable` has other, new features as well?

Piedone (changes requested):

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Psichorex: I don't quite get the point of you question now. Could you give more detail on what do you mean?
featuresToEnable can't change during 1 handler execution. It is just querying the features that should be enabled depending on the condition features. We needed the list field here to store features that the handler have already turned on because OC seems to be unable to get the enabled features during run time.
By this I mean that shellDescriptor.Features will not include the Features that has just been enabled by the Handler.
So lets say the handler #1 execution return 50 enabled features from the shellDescriptor.
Then it will enable OrchardCore.Twitter.
Then the handler runs again. Execution #2 the shellDescriptor will again return 50 as the count of enabled features eventhough Twitter was enabled and it should be 51.
Imo this is an OC bug of somekind. This is why I called this a workaround. This was causing the problem as the code could never actually check if the feature is turned on because the list of enabled features never changed. And it tried to enabled features twice which caused the invalid serial number for shell descriptor error.

Piedone: `featuresToEnable` can have more than one item, right? If indeed, then it's possible that e.g. item 1 will already be in `_previouslyAddedFeatures`, and item 2 won't. In this case, the method will return after item 1, but item 2 is something that should be enabled.

jtkech: For info the `...ing` handlers are first called, then the shell descriptor is updated (immediately committed internally), then the `...ed` handlers are called but in a deferred task so in its own scope (the handlers are scoped services).

I don't think that's a good pattern to call `.EnableFeaturesAsync()` in a `...ed` handler (called for each feature) because it triggers another shell descriptor update, a new deferred task in which the scoped `...ed` handlers are called again.

Would need to look at it more in details but it looks like a new dependency graph. For info a feature can be marked as `EnabledByDependencyOnly` (manifest attribute) meaning that it is auto enabled if at least a dependent feature is enabled, and kept enabled as long as one dependent feature is enabled.

But you may need to define it from configuration (not manifest files), not fully sure but as I remember we have filters/handlers to config feature infos and/or define custom dependency strategies.

Psichorex: @jtkech Thank you so much JT for sharing this it really helps. 
As for the `EnabledByDependecyOnly` part it would be good but this feature serves the purpose if one enables a condition feature then the conditional feature should be auto enabled but this does not mean that one could not directly enable the conditional feature.

Psichorex: @jtkech Do you have any idea why `EnableFeaturesAsync()` throws an `InvalidOperationException` error when:
image
So now I am really strictly trying to avoid feature enabling duplication.
Which works because I am trying to turn on 3 features:
Twitter, Facebook, Forms
Here are the condition and conditinals: 
image

Upon enabling Lombiq.ChartJs the feature turns on OrchardCore.Twitter and OrchardCore.Facebook and it works perfectly.

Then comes Lombiq.UIKit.
The handler is able to extract the feature that have already been enabled (Twitter and Facebook) and it returns OrchardCore.Forms on to the `featuresToEnable` collection.
Then `await.EnableFeaturesAsync(featuresToEnable)` containing only `OrchardCore.Forms` is throwing an error.
So basically not only is `await EnableFeaturesAsync()` throwing an exception when you try to enable a feature that has already been enabled but even if the feature is not enabled.
I don't understand why it works for the first execution when the condition is `Lombiq.ChartJs` and why it breaks at `Lombiq.UIKit`

Psichorex: P.S the whole above written story happens when I create a new Tenant from the Default tenant.
So this handler will run for every single feature that is enabled by default when creating a tenant.
It activates Lombiq.ChartJs first and after that some point it will activate Lombiq.UIKit.
When it activates Lombiq.UIKit the handler will fail with what I have described above

Piedone: This is the exception:

```
2023-05-06 14:47:16.9259|Test|0HMQE600KNIBJ||OrchardCore.Environment.Shell.ShellFeaturesManager|ERROR|IFeatureEventHandler thrown from Lombiq.Hosting.Tenants.FeaturesGuard.Handlers.FeaturesEventHandler by InvalidOperationException System.InvalidOperationException: Invalid serial number for shell descriptor
   at OrchardCore.Environment.Shell.Data.Descriptors.ShellDescriptorManager.UpdateShellDescriptorAsync(Int32 priorSerialNumber, IEnumerable`1 enabledFeatures)
   at OrchardCore.Environment.Shell.ShellDescriptorFeaturesManager.UpdateFeaturesAsync(ShellDescriptor shellDescriptor, IEnumerable`1 featuresToDisable, IEnumerable`1 featuresToEnable, Boolean force)
   at OrchardCore.Environment.Shell.ShellFeaturesManagerExtensions.EnableFeaturesAsync(IShellFeaturesManager shellFeaturesManager, IEnumerable`1 features, Boolean force)
   at Lombiq.Hosting.Tenants.FeaturesGuard.Handlers.FeaturesEventHandler.EnableConditionallyEnabledFeaturesAsync(IFeatureInfo featureInfo) in E:\ShortPath\OSOCE2\src\Modules\Lombiq.Hosting.Tenants\Lombiq.Hosting.Tenants.FeaturesGuard\Handlers\FeaturesEventHandler.cs:line 102
   at OrchardCore.Modules.InvokeExtensions.InvokeAsync[TEvents,T1](IEnumerable`1 events, Func`3 dispatch, T1 arg1, ILogger logger)    at OrchardCore.Environment.Shell.Data.Descriptors.ShellDescriptorManager.UpdateShellDescriptorAsync(Int32 priorSerialNumber, IEnumerable`1 enabledFeatures)
   at OrchardCore.Environment.Shell.ShellDescriptorFeaturesManager.UpdateFeaturesAsync(ShellDescriptor shellDescriptor, IEnumerable`1 featuresToDisable, IEnumerable`1 featuresToEnable, Boolean force)
   at OrchardCore.Environment.Shell.ShellFeaturesManagerExtensions.EnableFeaturesAsync(IShellFeaturesManager shellFeaturesManager, IEnumerable`1 features, Boolean force)
   at Lombiq.Hosting.Tenants.FeaturesGuard.Handlers.FeaturesEventHandler.EnableConditionallyEnabledFeaturesAsync(IFeatureInfo featureInfo) in E:\ShortPath\OSOCE2\src\Modules\Lombiq.Hosting.Tenants\Lombiq.Hosting.Tenants.FeaturesGuard\Handlers\FeaturesEventHandler.cs:line 102
   at OrchardCore.Modules.InvokeExtensions.InvokeAsync[TEvents,T1](IEnumerable`1 events, Func`3 dispatch, T1 arg1, ILogger logger)
```

As apparent from the stack trace, `FeaturesEventHandler` is called recursively, since the event is raised also after a feature is enabled from the event itself.

Try the `EnablingAsync` event instead, as JT above perhaps suggests.

Psichorex: EnablingAsync I have tried and it wasn't even hitting the breakpoint for me. But I will double check although I am quite sure this was the case

Piedone: If it's not usable, try to figure out how you can prevent this recursive enable, because I guess that's causing the issue.

Psichorex: At first I thought that the problem is due to the fact that if there are multiple conditions for 1 feature the Handler will try to enable the feature as many times and as the filtering for already enabled features is not working this was causing error.
I have implemented a filtering solution which , from debugging at least shows, prevents the handler to enable the same feature twice. After I have come up with this solution and I was testing it it turned out that even after filtering the `Invalid serial number for shell descriptor` issue is still present. I see from debugging that the collection passed to `EnableFeaturesAsnyc()` is clearly not containing the previous features just the one that is not yet enabled and the error still pops.

Psichorex: Okay so now I have debugged using `EnablingAsync` and the reason I was not hitting any breakpoints because my breakpoints were placed in a position that is really not hit. I put some at the begging and it seems that `EnablingAsync` is going through a handful of features only like really core OC features. I manually watched all the executions and features that are enabled by the recipe were not included. So I never hit the condition features.

Psichorex: Demo step by step:

Features I want to enable:
image

Here Facebook and Twitter is being enabled and the feature the Handler is hitting is Lombiq.ChartJs
image

Now the feature is Lombiq.UIKit. You can see the collection only has OrchardCore.Forms and the other 2 has successfully been removed from the collection and is ready to be passed to `EnableFeaturesAsync()`
image

Eventhough it clearly got a unique collection with a feature that hasn't been enabled yet the error occures:
image

Piedone: Try to understand the root cause of the exception by working back from there. You can use the debugger for that, see https://lombiq.atlassian.net/wiki/spaces/DEV/pages/1400143919/Debugging+a+NuGet-based+Orchard+Core+solution.

jtkech: @Psichorex Oops sorry I missed your last comments, I will try to look at it this night.

In the meantime if I understood correctly the logic, e.g. if `.UIKit` is enabled, `.Forms` should be enabled and can't be disabled, if `.UIKit` is disabled `.Forms` can be enabled or disabled, right?

For info, when we update the shell descriptor we use `GetOrCreateMutableAsync<>()` which returns a new descriptor instance (other updates will get this same instance) on which we increment the serial number. But when `ShellFeaturesManager` update again the shell descriptor it still use the serial number of its originally resolved `_shellDescriptor` (considered as immutable).

Here we could have also used `GetOrCreateMutableAsync<>()`, but when we introduced the usage of `IDocumentStore` with its own scoped caches we didn't change this behavior. I thought about it but make sense that you can't update the shell descriptor twice in the same shell scope.

Also in a `...ed` handler executed in a deferred task in its own scope, the shell descriptor has been already saved, we do this to only call the handler in that case.

jtkech: Okay, I looked at it, I suggest to use an `IShellDescriptorManagerEventHandler` that updates the shell descriptor, one problem is that `ShellDescriptorManager` does a `.CommitAsync()` before caliing the handlers, this is done because `ShellHost` (being currently the only one handler) releases the shell so that a subsequent request will build a new shell. So the `ShellDescriptor` needs to be committed before, otherwise `ShellHost` may register a shell that was built with a wrong shell descriptor.

So when we update the `ShellDescriptor` we need to commit again and release the shell again at the very end. Then we would need to call the features enable/disable handlers and so on, repeating what the existing services do.

So I opted to do it in a separated deferred task using its own scope, so as it would be done by another request but here initiated by the feature guarding, and then by using the existing services

Here a testing code where I only check 2 features but just to show the idea.

    .AddOrchardCms()
    .ConfigureServices(services =>
        services.AddScoped<IShellDescriptorManagerEventHandler, ShellFeaturesGuard>())


    public class ShellFeaturesGuard : IShellDescriptorManagerEventHandler
    {
        public Task ChangedAsync(ShellDescriptor descriptor, ShellSettings settings)
        {
            var conditions = new Dictionary<string, string[]>
            {
                ["OrchardCore.Feeds"] = new[] { "OrchardCore.Apis.GraphQL" }
            };

            var featureIdsToEnable = new HashSet<string>();
            foreach (var kvp in conditions.ToArray())
            {
                if (descriptor.Features.FirstOrDefault(f => f.Id == kvp.Key) is not null)
                {
                    continue;
                }

                foreach (var condition in kvp.Value)
                {
                    if (descriptor.Features.FirstOrDefault(f => f.Id == condition) is null)
                    {
                        continue;
                    }

                    featureIdsToEnable.Add(kvp.Key);

                    break;
                }
            }

            if (featureIdsToEnable.Count > 0)
            {
                ShellScope.AddDeferredTask(async scope =>
                {
                    var shellFeaturesManager =
                        scope.ServiceProvider.GetRequiredService<IShellFeaturesManager>();

                    var featuresToEnable = (await shellFeaturesManager
                        .GetAvailableFeaturesAsync())
                        .Where(feature => featureIdsToEnable.Contains(feature.Id))
                        .ToArray();

                    if (featuresToEnable.Length > 0)
                    {
                        await shellFeaturesManager.EnableFeaturesAsync(featuresToEnable, true);
                    }
                });
            }

            return Task.CompletedTask;
        }
    }

Psichorex: @jtkech Thank you JT! It seems that using `AddDeferredTask` did make the difference.

Piedone (approved):

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
jtkech: @Psichorex Okay good to know!

Yes, I was suggesting an `IShellDescriptorManagerEventHandler` but if it works by still using an `IFeatureEventHandler` that's okay. Take care about recursions that are still possible, for example if you need to both enable and disable the same feature because of a bad options configuration.

---

Hmm one drawback I see is that if we enable multiple features in one pass, we may trigger multiple deferred tasks. So one suggestion would be that in your scoped handler each time a Condition is enabled/disabled you trigger the deferred task but only once by using a local boolean.

Then do all the logic in the deferred task, first resolve the `ShellDescriptor`, yes better anyway to retrieve it from the deferred task to be sure of the current state, can be injected directly by using.

    scope.ServiceProvider.GetRequiredService<ShellDescriptor>()

Then lookup on your options dictionary, if a Conditional is or is not in the descriptor check if a Condition is enabled or not, and update features to enable/disable hashsets. Maybe also check that a feature is not in both collections, otherwise remove from one + maybe a log message.

Then update the features by using the resolved feature manager. The above could be done by a service injecting the descriptor and all it needs so that you only have to resolve this service, and it could also be the `IFeatureEventHandler` (not same scoped instance) that triggers the deferred task.

---

Looking at the code, it seems that when we disable all Conditions we also want to disable the Conditional, if so it's really like if the Conditions has a dependency on the Conditional and if the Conditional is EnabledByDependencyOnly. So just for info I tried the following that works.

    ocBuilder.ApplicationServices
        .AddSingleton<IFeatureBuilderEvents, FeatureGuardBuilderEvents>()


    public class FeatureGuardBuilderEvents : FeatureBuilderEvents
    {
        public override void Building(FeatureBuildingContext context)
        {
            if (context.FeatureId == "OrchardCore.Apis.GraphQL")
            {
                var dependencies = context.FeatureDependencyIds.ToHashSet();
                if (dependencies.Add("OrchardCore.Feeds"))
                {
                    context.FeatureDependencyIds = dependencies.ToArray();
                }
            }

            if (context.FeatureId == "OrchardCore.Feeds")
            {
                context.EnabledByDependencyOnly = true;
            }
        }
    }

The only concern is that it may introduce a circular feature dependency if the Conditional also depends on the Condition, for example if `Media` should enable `Azure.Media` that depends on `Media.Cache` that also depends on `Media`. It would not fail but the order of services registrations (as drivers / handlers) and then the order of executions may change.

jtkech:
@Psichorex Okay, see my https://github.com/Lombiq/Hosting-Tenants/pull/55#discussion_r1192907383

jtkech:
@Psichorex 

Finally I think you are right to use an `IFeatureEventHandler` because `...ed` event are executed in the deferred task initiated by the feature manager, so the one that we trigger will run after which is better.

For info in OC I suggested a PR to have a `FeatureEventHandler` class defining virtual methods so that you just have to override the ones you need, as we already have with the `ModularTenantEvents` class.

Then I think that on any feature enabled/disabled event, that can be related to a condition or conditional feature, you can trigger a deferred task but as said only once by using a local boolean e.g. `_deferredTask`. And in the deferred task you can resolve your options and the shell feature manager to apply all the logic.

So at least we have 2 deferred tasks, the one of the shell feature and ours, and as a maximum, if we enable/disable features, 2 deferred tasks again for a total of 4 deferred tasks.

So here what I quickly did for testing and to show the idea.

    public class FeatureGuardEventHandler : FeatureEventHandler
    {
        private bool _deferredTask;

        public override Task DisabledAsync(IFeatureInfo feature) => CheckFeatureConditionsAsync(feature);
        public override Task EnabledAsync(IFeatureInfo feature) => CheckFeatureConditionsAsync(feature);

        public Task CheckFeatureConditionsAsync(IFeatureInfo feature)
        {
            if (_deferredTask)
            {
                return Task.CompletedTask;
            }

            _deferredTask = true;

            ShellScope.AddDeferredTask(async scope =>
            {
                // For testing, in place of resolving here the options.
                Dictionary<string, List<string>> conditions = new()
                {
                    { "OrchardCore.Feeds", new List<string> { "OrchardCore.Apis.GraphQL" } },
                };

                var shellFeaturesManager = ShellScope.Services.GetRequiredService<IShellFeaturesManager>();
                var enabledFeatures = (await shellFeaturesManager.GetEnabledFeaturesAsync()).ToArray();
                var enabledFeaturesIds = enabledFeatures.Select(f => f.Id).ToHashSet();

                var featuresToEnableIds = new HashSet<string>();
                var featuresToDisableIds = new HashSet<string>();

                foreach (var condition in conditions)
                {
                    var hasConditional = enabledFeaturesIds.Contains(condition.Key);
                    var hasCondition = enabledFeaturesIds.Intersect(condition.Value).Any();

                    if (hasCondition && !hasConditional)
                    {
                        featuresToEnableIds.Add(condition.Key);
                    }

                    if (!hasCondition && hasConditional)
                    {
                        featuresToDisableIds.Add(condition.Key);
                    }
                }

                if (featuresToEnableIds.Count == 0 && featuresToDisableIds.Count == 0)
                {
                    return;
                }

                var avalaibleFeatures = await shellFeaturesManager.GetAvailableFeaturesAsync();

                var featuresToEnable = avalaibleFeatures
                    .Where(feature => featuresToEnableIds.Contains(feature.Id))
                    .ToArray();

                var featuresToDisable = enabledFeatures
                    .Where(feature => featuresToDisableIds.Contains(feature.Id))
                    .ToArray();

                if (featuresToEnable.Length == 0 && featuresToDisable.Length == 0)
                {
                    return;
                }

                await shellFeaturesManager.UpdateFeaturesAsync(featuresToDisable, featuresToEnable, true);
            });

            return Task.CompletedTask;
        }
    }

Notice that we no longer use `_shellDescriptorManager`, we could have and in that case, as said we can resolve the descriptor directly, for info `.GetShellDescriptorAsync()` is only useful to get the descriptor from the database to first build the tenant container, it is then registered in the tenant container DI.

Psichorex:
@jtkech Thanks for all the information. 
Based on my test after introducing the `AddDefferedTask()` part most of the issues are gone.
Also I made changes on how the service actually handles activation duplication:
image
In my opinion this is the safest way to never try to enable the same Feature twice using `LINQ.Except()`
This way allthough the `conditionalFeatureIds` might have already enabled features in it beause some of the condition features for them are the same, `featureIdsToEnable` will now never contain enabled features thanks to `Except()`
If the collection result of `Except()` has a Count of 0 we return as we do not have any Features that should be enabled.
I also introduced a part where we check for conditionals that do not exist:
image

This might help the usage of this whole service on condition there is a slight misspelling that would otherwise be hard to find.

Everything as far I have tested is working now.
Any combination of conditions and conditionals do not break the application.
Manually disabling conditions will disable conditionals we have a separate handler method that executes on `DisabledAsync...`
Manually turning back conditions will again turn on conditionals as intended.

jtkech:
@Psichorex All good then!

- Personally I would not keep `KeepConditionallyEnabledFeaturesEnabledAsync` and `DisableConditionallyEnabledFeaturesAsync`, just trigger the deferred task on any feature change (enabled / disabled) in your handler, but only once by using a local boolean e.g. `_deferredTaskRegistered`, so nothing to do if already set to true.

    Here we are multiplexing multiple feature events in only one, that's why I first suggested to use an `IShellDescriptorManagerEventHandler` but the event happens a little too early.

- And then apply all the logic in the deferred task (acting as a checker) based on the current state of the shell descriptor (that may have been changed in the meantime by another thread/process), this to change the state of any conditional feature if needed.

    Here we are handling the event in the deferred task.

Psichorex:
@jtkech I would rather avoid removing `KeepConditionallyEnabledFeaturesEnabledAsync` and `DisableConditionallyEnabledFeaturesAsync` as they both work and somewhat contribute to a more readable code as the functionalities are separated. Also they don't need to utilize a Deferred Task as disabling can only happen by user interaction which will only trigger 1 Handler execution and problem was only during a Tenant setup where all the basic OrchardFeatures and recipe enabled features were getting enabled continuously.

Also I tried the local boolean prior you mentioning it because I thought about that already but it is causing errors.
The first handler execution will set the bool to true and the DeferredTask will never run again.
Seems that every single Handler run has a different scope to it. So the DeferredTask will run as many times as the Handler but at least it prevents recursive calls which were causing the ShellDescriptor going out of sync.

If you think that the current solution is sufficient please leave an approved label on the PR as you have highly contributed on this review for which I am really grateful.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -54,56 +51,63 @@ async Task IFeatureEventHandler.DisabledAsync(IFeatureInfo feature)
     /// Enables conditional features (key) if one of their corresponding condition features (value) was enabled.
     /// </summary>
     /// <param name="featureInfo">The feature that was just enabled.</param>
-    public async Task EnableConditionallyEnabledFeaturesAsync(IFeatureInfo featureInfo)
+    public Task EnableConditionallyEnabledFeaturesAsync(IFeatureInfo featureInfo)
     {
-        if (_shellSettings.IsDefaultShell() ||
-            _conditionallyEnabledFeaturesOptions.Value.EnableFeatureIfOtherFeatureIsEnabled is not { } conditionallyEnabledFeatures)
+        // We are using a Deferred Task so we can avoid recursive executions upon tenant setup.
+        ShellScope.AddDeferredTask(async scope =>
```
Piedone, line 47: Shouldn't `DisableConditionallyEnabledFeaturesAsync` use a deferred task as well?

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -189,4 +187,32 @@ public async Task DisableConditionallyEnabledFeaturesAsync(IFeatureInfo featureI
...
+        {
+            conditionalFeatures = null;
+            return false;
+        }
+
+        conditionalFeatures = conditionallyEnabledFeatures;
+        return true;
+    }
+
+    private static IEnumerable<string> GetAllConditionFeatureIds(
```
Piedone: Here we're retrieving a distinct list of features that are to be conditionally enabled when `featureInfo` was enabled, right? If indeed, then apart from the name change also add this as docs on the method.

```suggestion
    private static IEnumerable<string> GetAllConditionalFeatureIds(
```

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -54,56 +51,63 @@ async Task IFeatureEventHandler.DisabledAsync(IFeatureInfo feature)
...
+            {
+                return;
+            }
 
-        var allConditionFeatureIds = new List<string>();
-        foreach (var conditionFeatureIds in conditionallyEnabledFeatures.Values)
-        {
-            allConditionFeatureIds.AddRange(conditionFeatureIds);
-        }
+            var allConditionFeatureIds = GetAllConditionFeatureIds(conditionallyEnabledFeatures, featureInfo);
```
Piedone: Why are we not just enabling all features from `allConditionFeatureIds` after this point? I don't get why we need all the other logic below (except for not trying to enable features that are already enabled, but that's something that should be handled by `IShellFeaturesManager` anyway).

Piedone (changes requested):

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Psichorex: No this is just a method to DRY the part where the condition features were collected in a new `var allConditionFeatureIds = new List<string>();`. So this is collecting the condition features not the conditional features. A conditinal feature would be `dictionary.Key` and not `dictionary.Value`

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Psichorex: Because the **Condition** features are the features that define whether the **Conditional** feature should be enabled or not. The whole Handler logic is to determine which **Conditional** feature to enable depending on the `featureInfo` if it is a condition feature.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Psichorex: In my opinion it doesn't need to use `DeferredTask` because disabling features never happens continuously N times after each other while Enabling features do when a Tenant is being initialized.
Disabling is mostly manually user controlled which was never broken before.
The only broken thing was the massive `EnabledAsync` calls when you create a new Tenant.
So unless there isn't a service that is mass disabling features one after another we don't actually need `deferredTask`

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Psichorex: Logic is the following
If `featureInfo` is a **condition** feature then we want to get all of the **conditional** features that depend on the current `featureInfo`. We want to check not to enable **conditionals** twice.  Also `allConditionFeatureIds` are strings so we would either way need to get their corresponding `FeatureInfo`. But again it's not **Condition** features we want to enable but **Conditionals**

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Psichorex: For safety measures we can wrap that with `DeferredTask` but I haven't experienced malfunctionalities without it.

Piedone: I see, then please add a comment into `DisableConditionallyEnabledFeaturesAsync` about "In my opinion it doesn't need to use DeferredTask because disabling features never happens continuously N times after each other while Enabling features do when a Tenant is being initialized."

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Piedone: I see, thanks.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Piedone: I see, thanks for explaining.

Piedone (approved):

jtkech:
@Psichorex No problem and sorry if I bother you a little ;) I only describe how i think it should be done.

>  ... they both work and somewhat contribute to a more readable code as the functionalities are separated.

For me there is only one feature, change the state of conditionals based on the state of their conditions. In the sample code (see below) I introduced some checkings, e.g you can't enable a `DefaultTenantOnly` feature, you can't disable an `IsAlwaysEnabled` feature (it would result in an infinite loop), I may have missed one but it's more maintainable to have all this logic in only one place.

> Also they don't need to utilize a Deferred Task as disabling can only happen by user interaction which will only trigger 1 Handler execution.

This is not a good reason, it should also work with a recipe (a setup one or not) disabling multiple features, anyway disabling a single feature may disable multiple ones through the dependency graph.

> Seems that every single Handler run has a different scope to it. So the DeferredTask will run as many times as the Handler.

As I have tried it doesn't behave like this, in a given scope by using a boolean our deferred task is triggered only once. But yes, if it enables/disables feature(s), our handler is called again in another deferred task (and scope) with the boolean equal to false, so we trigger our deferred task a second (and last) time but normally for nothing (unless another thread/process changed the features states).

But yes, as it is currently done without the bool, a deferred task is triggered for each feature to enable.

---

Also for info, yes in the deferred task you should resolve again scoped services as we are in a different scope. Yes, you can reference a tenant singleton, but in that case better to init a variable in the handler `var someService = _someService` so that the deferred task closure references that variable in place of the scoped handler holding `_someService`, this from another scope. For the options I recommend to resolve them in the deferred task too.

---

So I think it is worth to try the following where I introduced some additional checkings as mentioned above, quickly done so I could have made a mistake, otherwise if it doesn't work it would mean that there is something else to understand and fix.

    public class FeatureGuardEventHandler : FeatureEventHandler
    {
        private bool _deferredTask;

        public override Task DisabledAsync(IFeatureInfo feature) => TriggerDeferredTaskAsync();
        public override Task EnabledAsync(IFeatureInfo feature) => TriggerDeferredTaskAsync();

        public Task TriggerDeferredTaskAsync()
        {
            if (_deferredTask)
            {
                return Task.CompletedTask;
            }

            _deferredTask = true;

            ShellScope.AddDeferredTask(async scope =>
            {
                // Can be checked from the handler.
                if (scope.ShellContext.Settings.IsDefaultShell())
                {
                    //return;
                }

                var options = scope.ServiceProvider.GetRequiredService<IOptions<ConditionallyEnabledFeaturesOptions>>().Value;
                if (options.EnableFeatureIfOtherFeatureIsEnabled is null ||
                    options.EnableFeatureIfOtherFeatureIsEnabled.Count == 0)
                {
                    return;
                }

                var conditions = options.EnableFeatureIfOtherFeatureIsEnabled;
                var shellFeaturesManager = scope.ServiceProvider.GetRequiredService<IShellFeaturesManager>();

                var enabledFeatures = (await shellFeaturesManager.GetEnabledFeaturesAsync()).ToArray();
                var enabledFeaturesIds = enabledFeatures.Select(f => f.Id).ToHashSet();

                var featuresToEnableIds = new HashSet<string>();
                var featuresToDisableIds = new HashSet<string>();
                foreach (var condition in conditions)
                {
                    var hasConditional = enabledFeaturesIds.Contains(condition.Key);
                    var hasCondition = enabledFeaturesIds.Intersect(condition.Value).Any();

                    if (hasCondition && !hasConditional)
                    {
                        featuresToEnableIds.Add(condition.Key);
                    }

                    if (!hasCondition && hasConditional)
                    {
                        featuresToDisableIds.Add(condition.Key);
                    }
                }

                if (featuresToEnableIds.Count == 0 && featuresToDisableIds.Count == 0)
                {
                    return;
                }

                var avalaibleFeatures = await shellFeaturesManager.GetAvailableFeaturesAsync();
                var featuresToEnable = avalaibleFeatures
                    .Where(feature => featuresToEnableIds.Contains(feature.Id))
                    .ToArray();

                var featuresToDisable = enabledFeatures
                    .Where(feature => featuresToDisableIds.Contains(feature.Id))
                    .ToArray();

                if (featuresToEnable.Any(feature => feature.DefaultTenantOnly || feature.EnabledByDependencyOnly))
                {
                    throw new InvalidOperationException("Can't enable a 'DefaultTenantOnly' feature ...");
                }

                if (featuresToDisable.Any(feature => feature.IsAlwaysEnabled || feature.EnabledByDependencyOnly))
                {
                    throw new InvalidOperationException("Can't disable an 'IsAlwaysEnabled' feature ...");
                }

                if (featuresToEnable.Length == 0 && featuresToDisable.Length == 0)
                {
                    return;
                }

                if (featuresToEnable.Intersect(featuresToDisable).Any())
                {
                    throw new InvalidOperationException("A feature can't be both enabled and disabled ...");
                }

                await shellFeaturesManager.UpdateFeaturesAsync(featuresToDisable, featuresToEnable, true);
            });

            return Task.CompletedTask;
        }
    }

Psichorex:
@jtkech I reworked the whole Handler based on your solution.
You are right and now I see it's much more clear to have 1 Handling method for all of this. I refactored your solution so that it fits into our analysers. But now there's only one method in a `DeferredTask` and everything is working.
@Piedone Please review the new code here.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -13,215 +12,118 @@ namespace Lombiq.Hosting.Tenants.FeaturesGuard.Handlers;
...
     private readonly IOptions<ConditionallyEnabledFeaturesOptions> _conditionallyEnabledFeaturesOptions;
-    private readonly ShellSettings _shellSettings;
 
-    public FeaturesEventHandler(
-        IShellFeaturesManager shellFeaturesManager,
-        IOptions<ConditionallyEnabledFeaturesOptions> conditionallyEnabledFeaturesOptions,
-        ShellSettings shellSettings)
-    {
-        _shellFeaturesManager = shellFeaturesManager;
+    public FeaturesEventHandler(IOptions<ConditionallyEnabledFeaturesOptions> conditionallyEnabledFeaturesOptions) =>
```
Piedone: I'd resolve this from inside the deferred task too, then.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -13,215 +12,118 @@ namespace Lombiq.Hosting.Tenants.FeaturesGuard.Handlers;
...
     Task IFeatureEventHandler.UninstalledAsync(IFeatureInfo feature) => Task.CompletedTask;
 
     /// <summary>
-    /// Enables conditional features (key) if one of their corresponding condition features (value) was enabled.
+    /// Enables or disables conditional features depending on ConditionallyEnabledFeaturesOptions.
+    /// Prevents disabling features that should be enabled according to their conditions.
     /// </summary>
-    /// <param name="featureInfo">The feature that was just enabled.</param>
-    public Task EnableConditionallyEnabledFeaturesAsync(IFeatureInfo featureInfo)
+    private Task HandleConditionallyEnabledFeaturesAsync()
```
Piedone, line 38: Much better!

Piedone (changes requested):

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Psichorex: This is actually used in the helper method which doesn't have a scope instance.
image

If you mean to resolve this in the deferredTask and then pass it as a parameter to the `GetFeaturesToBeEnabledAndDisabled` then it's okay.

Psichorex: But resolving it in the helper method would require another service to get service which kills the point imo.

Piedone: You can pass the resolved options to the method, yes.

jtkech:
@Psichorex Thanks for following my suggestions

One remaining important thing, try the usage of the boolean, when I tried it on a setup recipe that enables let's say 10 features, I could see that the deferred task is trigerred once in place of 10 times.

        if (_deferredTask)
        {
            return Task.CompletedTask;
        }

        _deferredTask = true;

        ShellScope.AddDeferredTask(async scope =>

---

Also, as I did in my code sample, I think we should check if a conditional feature is marked as `DefaultTenantOnly`, `IsAlwaysEnabled` or `EnabledByDependencyOnly`, for example you will never be able to disable an `IsAlwaysEnabled` feature which would end up in an infinite loop.

---

The options `.Value` can be get in the ctor to set a local `ConditionallyEnabledFeaturesOptions` variable, this in place of using an `IOptions<ConditionallyEnabledFeaturesOptions>` type. But anyway as @Piedone suggested better to resolve it in the deferred task too and then pass its value to your method.

Hmm, you can also pass the dictionary property directly, as you want, but in any case I don't like the following check using ` is not { }`, but maybe just because I'm not used to see it ;)

    if (options.EnableFeatureIfOtherFeatureIsEnabled is not { } conditionallyEnabledFeatures)

Psichorex:
@jtkech I added some checks to cover those scenarios.

`is not { }` is no longer in the code as we pass the `IOptions<ConditionallyEnabledFeaturesOptions>().Value` as a param so now the check is `if (!dictionary.Any())` which is probably a bit more readable although I really like to use `is { } variable` shorthands as they really save some lines.
Thank you for all your help I think we really got this FeaturesGuard module to be much better and cleaner.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -14,168 +13,137 @@ namespace Lombiq.Hosting.Tenants.FeaturesGuard.Handlers;
...
+            if (featuresToEnable.Any(feature => feature.DefaultTenantOnly || feature.EnabledByDependencyOnly))
+            {
+                throw new InvalidOperationException("'DefaultTenantOnly' feature can't be enabled by FeaturesGuard");
+            }
+
+            var featuresToDisable = enabledFeatures
+                .Where(feature => featuresToDisableIds.Contains(feature.Id))
+                .ToList();
+
+            if (featuresToEnable.Any(feature => feature.IsAlwaysEnabled || feature.EnabledByDependencyOnly))
```
jtkech: Here you should use `featuresToDisable`.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -14,168 +13,137 @@ namespace Lombiq.Hosting.Tenants.FeaturesGuard.Handlers;
...
-        }
-
-        var allConditionFeatureIds = new List<string>();
-        foreach (var conditionFeatureIds in conditionallyEnabledFeatures.Values)
-        {
-            allConditionFeatureIds.AddRange(conditionFeatureIds);
-        }
-
-        if (!allConditionFeatureIds.Contains(featureInfo.Id))
+        ShellScope.AddDeferredTask(async scope =>
```
jtkech, line 47: Just before, try to use and check a `_deferredTask` boolean as suggested.

It prevents many deferred tasks from being triggered.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -108,7 +108,7 @@ private static Task HandleConditionallyEnabledFeaturesAsync()
     /// <returns>A boolean value whether ConditionallyEnabledFeaturesOptions is populated or not.
     /// Also produces <paramref name="featuresToEnable"/> and <paramref name="featuresToDisable"/>.
     /// </returns>
-    private static bool GetFeaturesToBeEnabledAndDisabled(
+    private static bool TryGetFeaturesToBeEnabledAndDisabled(
```
Piedone, line 120: Methods returning `bool` and having `out` parameters follow this convention in .NET.

Piedone (changes requested):
Address the new remarks of JT, please.

jtkech:
@Psichorex Okay no problem, thanks.

Just did a last little review ;)

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Psichorex: @jtkech I tried the `_deferredTaskExecuted` boolean like you have suggested but there is a catch for me with it.
So it basically prevents the handler to run multiple times that is understandable but this for me kills the functionality of the half feature.
For example `OrchardCore.Facebook` and `OrchardCore.Twitter` will be enabled if `Lombiq.UIKit` is enabled.
But after this point if I disable `Lombiq.UIKit` the other 2 features will remain enabled. Also If I want to turn off Facebook I can turn it off eventhough it should stay enabled because `Lombiq.UIKit` is still enabled.
But the handler is not running once more because of the boolean set to true.

Psichorex:
@Piedone What is the ideal functionality of this module according to you?
1)
- If a condition is enabled -> enable the corresponding conditional and keep it enabled as long as the condition is enabled, also if you disabled the condition the conditional should get disabled aswell. (**This way we need to avoid using local boolean**)
2)
- We use a local boolean to save executions in the handler but as far as I have tested this boolean will result in preventing the handler to run even after tenant setup. So basically after tenant setup if I want to disable the conditional I can do that eventhough its condition is still on. Also if I turn off all the conditions the conditionals remain enabled because again the handler never run again.

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Piedone: This delegate handles any number of features to be enabled/disabled, no? So it shouldn't require running more than once for an updated shell.

Piedone:
The desired functionality is option 1, i.e. what happens today.

Piedone (approved):

jtkech:
@Psichorex 

> For example `OrchardCore.Facebook` and `OrchardCore.Twitter` will be enabled if `Lombiq.UIKit` is enabled. But after this point if I disable `Lombiq.UIKit` the other 2 features will remain enabled.

Where and how you will then disable `Lombiq.UIKit`? If through the feature manager, it will trigger another deferred task in which our handler will be resolved again but from a different scope, so normally the boolean is no longer equal to true, unless you registered your handler as a tenant singleton.

As @Piedone said

> So it shouldn't require running more than once for an updated shell.

Otherwise there is something else to understand, do you have a repro that I could try?

Otherwise the other parts of the code LGTM.

jtkech:
@Psichorex 

First, by using the shell feature manager we can't enable/disable a set of features twice in the same shell scope, otherwise we will get an "Invalid Shell Descriptor Serial Number".

So, I did the following test in the Features controller, when we disable a condition feature, we trigger 2 nested deferred task to re-enable it and then re-disable it (tried the same when enabling it), it works well on my side, both my condition and conditional features are in the right state at the end.

I could see the boolean set to true preventing from triggering other deferred tasks, then returning back to false (meaning the handler is resolved from another scope), then set again to true.

To declare a `_deferredTask` boolean and use it before triggering the deferred task in your method, you need to remove the `static` modifier from your handler method.

For the test I only added and used the above boolean in the handler, the rest being copy pasted from your code that seems to work well. Here the code of my test added to the Features Controller.

    if (id == "OrchardCore.Apis.GraphQL")
    {
        ShellScope.AddDeferredTask(async scope =>
        {
            var shellFeaturesManager = scope.ServiceProvider
                .GetRequiredService<IShellFeaturesManager>();

            await shellFeaturesManager.EnableFeaturesAsync(new[] { feature }, true);
            ShellScope.AddDeferredTask(async scope =>
            {
                var shellFeaturesManager = scope.ServiceProvider
                    .GetRequiredService<IShellFeaturesManager>();

                await shellFeaturesManager.DisableFeaturesAsync(new[] { feature }, true);
            });
        });
    }

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
```diff
@@ -14,168 +13,136 @@ namespace Lombiq.Hosting.Tenants.FeaturesGuard.Handlers;
...
     Task IFeatureEventHandler.InstallingAsync(IFeatureInfo feature) => Task.CompletedTask;
 
     Task IFeatureEventHandler.InstalledAsync(IFeatureInfo feature) => Task.CompletedTask;
 
     Task IFeatureEventHandler.EnablingAsync(IFeatureInfo feature) => Task.CompletedTask;
 
-    Task IFeatureEventHandler.EnabledAsync(IFeatureInfo feature) => EnableConditionallyEnabledFeaturesAsync(feature);
+    Task IFeatureEventHandler.EnabledAsync(IFeatureInfo feature) => HandleConditionallyEnabledFeaturesAsync();
 
     Task IFeatureEventHandler.DisablingAsync(IFeatureInfo feature) => Task.CompletedTask;
```
jtkech: For all methods no need to use the interface name, just add the public modifier, for example

`Task IFeatureEventHandler.EnabledAsync()` can be replaced by `public Task EnabledAsync()`

Piedone (changes requested):
Please address JT's remarks.

jtkech:
Yes, using a local bool (to reduce the deferred tasks) works well on my side (if the handler is a scoped service), if @Psichorex has an use case that doesn't work we would need to understand what happens.

    if (_deferredTask)
    {
        return Task.CompletedTask;
    }

    _deferredTask = true;

    ShellScope.AddDeferredTask(async scope =>

Lombiq.Hosting.Tenants.FeaturesGuard/Handlers/FeaturesEventHandler.cs
Psichorex: @Piedone The problem if it runs once than as I described 

> Yes, using a local bool (to reduce the deferred tasks) works well on my side (if the handler is a scoped service), if @Psichorex has an use case that doesn't work we would need to understand what happens.
> 
> ```
> if (_deferredTask)
> {
>     return Task.CompletedTask;
> }
> 
> _deferredTask = true;
> 
> ShellScope.AddDeferredTask(async scope =>
> ```

@jtkech Here is my repro if I use `_deferredTask` boolean:
1) Create a new tenant from the default tenant
2) Setup the new tenant
3) When clicking setup the Handler will run and it will enable all conditional features and `_deferredTask` will be true.
4) Go to the new Tenant admin and try to disable a condition feature or a conditional feature
5) The handler for me never runs again thus I can disable a conditional feature eventhough it's condition is still on and it should be kept enabled.

jtkech: Thanks for the repro, I will try it this night.

Can you show me the code that registers the handler and defines the `_deferredTask` variable.

Psichorex: @jtkech Sending you what you need.

image
image
image

Psichorex: For me this is preventing the handler to ever be executed again.
So basically this will execute only once when the tenant is being initialized.

jtkech: Okay just retried, I can't repro, it works well on my side, `_deferredTask` is false again when the handler is resolved from another scope. Did you retry it with the new code where states are retrieved and all the logic done in the deferred task?

I used the same code unless that I always configure the options by code (one conditional/condition) and that I always register the handler by code, looks like you are doing it through a module/feature startup.

Okay, just retried by using a separate `OC.FeaturesGuard` that I included in the setup recipe, all still works fine, the only difference is that I still always config my testing options by code (not from the configuration) in the module startup.

As a side note, not tried but I think you could get rid of the `ConditionallyEnabledFeatures` config section and then bind directly on `ConditionallyEnabledFeaturesOptions`, but as you want.

jtkech:
@Psichorex

Nothing urgent, just to be sure that you read my last comments.

https://github.com/Lombiq/Hosting-Tenants/pull/55/files#r1201022456
https://github.com/Lombiq/Hosting-Tenants/pull/55#issuecomment-1556452004

All works fine for me, so I'm still interested to know what happens on your side ;)

jtkech:
@Piedone @Psichorex Any feedback on my last comments/suggestions?

jtkech:
@Piedone did you get notified about my last comments/suggestions??

Piedone:
I got it, yes, but it's up to Roland to reply.

Psichorex:
For me doing anything in Feature Manager never executes the DefferedTask again. So locally the functionality for me with a local boolean is nit working as desired per Piedone. I will double check it but I was not able to catch any breakpoints after the tenant was set up.

jtkech:
> I will double check it

Yes please.

> but I was not able to catch any breakpoints after the tenant was set up.

That's the weird thing for me ;)

How a boolean held by a scoped service can be then always equal to true?

If it still doesn't work for you, can you share a simple project and repro steps?

So that I could check it with the exact same code, thanks.

Piedone (changes requested):
Address the remaining remarks of JT.

Psichorex:
@jtkech Hi JT I found the problem was caused by Rider suggesting that the boolean should be `private static bool` but making the field static was causing it to keep it's value eventhough we are shifting scoped.
Now everything is working.

Piedone (approved):

Psichorex:
@Piedone That `.Exists()` in `Lombiq.Hosting.Tenants.Management/Services/SetupWithRecipesFilterService` call is failing the NuGet Publish for some reason. It is very interesting because it is failing with a build error but I can't see any. Replacing the `Exists` with `Any()` could be a quick solution.

Piedone:
It's just that Helpful Libraries got an update since the most recent NuGet alpha, so I'm publishing a new package of it too.

jtkech:
@Psichorex 

Okay cool then.

Only one thing, you set the boolean to true in the wrong place, should be set before triggering the deferred task, not inside the deferred task itself, this is not an `Executed` but a `Triggered` checking.

        if (_deferredTaskTriggered)
        {
            return Task.CompletedTask;
        }

        _deferredTaskTriggered = true;
        ShellScope.AddDeferredTask(async scope =>
        {
            ...

Otherwise deferred tasks would be still triggered, anyway not so good to reference a field of a service resolved from a scope that is no longer valid, better to use an intermediate var but normally for read only.

Psichorex:
@jtkech Thank you so much for all the help and knowledge you gave here. I really appriciate it. 
I also fixed your last remark.

Piedone (approved):

jtkech:
@Psichorex no problem, thanks for the last fix

PR #60, opened 2023-05-18
OSOE-621: Feature for adding noindex, nofollow meta tags in non-production in Lombiq.Hosting.Tenants

### LINKED ISSUE #56: Feature for adding noindex, nofollow meta tags in non-production (OSOE-621)
Staging and other non-production environments can be open to the public internet. To prevent search bots from indexing these (and thus duplicating production content), let's add a new module that when enabled, adds a `noindex, nofollow` meta tag when the app is not running in a production environment.

```html
<meta name="robots" content="noindex, nofollow">
```

The environment can be checked with `IHostEnvironment.IsProduction()`. However, add a separate configuration coming from a configuration provider that indicates if the current environment is production or non-production, in case ASP.NET Core should be told it runs in Production but not actually (like in staging, that should be as close to production as possible).

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-621)

DemeSzabolcs:
[OSOE-621](https://lombiq.atlassian.net/browse/OSOE-621)
Fixes #56

[OSOE-621]: https://lombiq.atlassian.net/browse/OSOE-621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Hosting.Tenants.EnvironmentRobots.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+    public static async Task TestRobotMetaTagIsMissingAsync(this UITestContext context, bool shouldBeMissing)
+    {
+        var metaTagXPath = By.XPath($"//meta[@name='robots' and @content='noindex, nofollow']").OfAnyVisibility();
+
+        await context.SignInDirectlyAsync();
+        await context.GoToHomePageAsync();
+
+        // Checking the response header with JavaScript.
+        var isHeaderPresent = context.Driver.ExecuteAsyncScript(@"
+            var callback = arguments[arguments.length - 1];
```
Psichorex: I know it's an inline script therefore it won't get analysed but with ES6 we don't use `var` in JS anymore.
Please utilize `const` and `let` according to the variable's scope.

Lombiq.Hosting.Tenants.EnvironmentRobots.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+            var xhr = new XMLHttpRequest();
+            xhr.open('GET', window.location.href);
+            xhr.send();
+            xhr.onload = function() {
+                var xRobotsTag = xhr.getResponseHeader('X-Robots-Tag') ?? '';
+                callback(xRobotsTag.includes ('noindex, nofollow'));
```
Psichorex: Please remove the space between includes function and its parameters to improve readability.

Lombiq.Hosting.Tenants.EnvironmentRobots/Extensions/IHostEnvironmentEnvironmentRobotsExtensions.cs
```diff
@@ -0,0 +1,12 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+
+namespace Lombiq.Hosting.Tenants.EnvironmentRobots.Extensions;
+public static class IHostEnvironmentEnvironmentRobotsExtensions
+{
+    public static bool IsProductionWithConfiguration(
+        this IHostEnvironment hostEnvironment,
+        IOptions<EnvironmentRobotsOptions> options) =>
+        options.Value.IsProduction ?? hostEnvironment.IsProduction();
```
Psichorex, line 11: This line should be +1 indent compared to parameters. At least if I remember correctly.

Lombiq.Hosting.Tenants.EnvironmentRobots/Filters/EnvironmentRobotsMetaTagFilter.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                Name = "robots",
+                Content = "noindex, nofollow",
+            });
+        }
+    }
+
+    public void OnResultExecuted(ResultExecutedContext context)
+    {
+        // Intentionally left empty, we don't need the function, but we need to add it because of the interface
```
Psichorex: My suggestion would be :
"Intentionally empty. Required by interfance implementation only." to be more formal.
@Piedone It is really often that we need empty methods. We could add a custom attribute to one of our libraries that handles these situation. Something like:
`[PlaceholderImplementation("Interface requirement")]` 
this way we don't need comments anymore.

Lombiq.Hosting.Tenants.EnvironmentRobots/Middlewares/EnvironmentRobotsMiddleware.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    public Task InvokeAsync(HttpContext context)
+    {
+        if (!_hostEnvironment.IsProductionWithConfiguration(_options))
+        {
+            var headerValue = context.Response.Headers["X-Robots-Tag"].FirstOrDefault() ?? string.Empty;
+
+            var directives = new List<string>();
+
+            if (!headerValue.Contains("noindex")) directives.Add("noindex");
```
Psichorex: Avoid one-liner if statements when there are also ifs with `{ }`. 
Keep it consistent in my opinion is the best for readability.

Lombiq.Hosting.Tenants.EnvironmentRobots/Models/EnvironmentRobotsOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+/// </summary>
+public class EnvironmentRobotsOptions
+{
+    /// <summary>
+    /// Gets or sets a value indicating whether to add a meta tag and response header with noindex, nofollow values when
+    /// the app is not running in a production environment. It adds the tags when <see cref="IsProduction"/> is <see
+    /// langword="false"/>. When set it overrides <see cref="HostEnvironmentEnvExtensions.IsProduction"/>'s result,
+    /// which the module is using by default.
+    /// </summary>
+    public bool? IsProduction { get; set; }
```
Psichorex, line 16: I don't think this needs to be nullable.
At least we did not use any nullable properties in most of our `Options` classes.

Psichorex (changes requested):

Lombiq.Hosting.Tenants.EnvironmentRobots/Models/EnvironmentRobotsOptions.cs
DemeSzabolcs: It should be nullable, so by default it can use `IHostEnvironment.IsProduction()` (if it's null) so we don't need to configure it unless:
> in case ASP.NET Core should be told it runs in Production but not actually (like in staging, that should be as close to production as possible).

Psichorex: I see thanks for clarifying. This is resolved in this case.

Lombiq.Hosting.Tenants.EnvironmentRobots.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+public static class TestCaseUITestContextExtensions
+{
+    public static async Task TestRobotMetaTagIsMissingAsync(this UITestContext context, bool shouldBeMissing)
+    {
+        var metaTagXPath = By.XPath($"//meta[@name='robots' and @content='noindex, nofollow']").OfAnyVisibility();
+
+        await context.SignInDirectlyAsync();
+        await context.GoToHomePageAsync();
+
+        // Checking the response header with JavaScript.
```
Piedone: This is apparent from the code, but explain _why_ do we use JS.

Lombiq.Hosting.Tenants.EnvironmentRobots.Tests.UI/Readme.md
```diff
@@ -0,0 +1,7 @@
+# Lombiq Hosting - Tenants EnvironmentRobots for Orchard Core - UI Test Extensions
+
+## About
+
+Extension methods that tests features in Lombiq Hosting - Tenants EnvironmentRobots, with the help of [Lombiq UI Testing Toolbox for Orchard Core](https://github.com/Lombiq/UI-Testing-Toolbox).
```
0liver: Typo:
```suggestion
Extension methods that test features in Lombiq Hosting - Tenants EnvironmentRobots, with the help of [Lombiq UI Testing Toolbox for Orchard Core](https://github.com/Lombiq/UI-Testing-Toolbox).
```

Lombiq.Hosting.Tenants.EnvironmentRobots/Extensions/IHostEnvironmentEnvironmentRobotsExtensions.cs
0liver: No auto-formatter should inline it, so please don't, because the next time someone will format the file, this line will change.

It uses the same indentation as a multi-statement method body like this:

image

Lombiq.Hosting.Tenants.EnvironmentRobots/Filters/EnvironmentRobotsMetaTagFilter.cs
0liver: Alternatively, we could provide a base class with virtual empty methods which would contain the suppression, derive from that, and only override the method needed. That would save us from adding the suppression altogether.

Lombiq.Hosting.Tenants.EnvironmentRobots/Extensions/IHostEnvironmentEnvironmentRobotsExtensions.cs
0liver, line 11: `EnvironmentRobotsOptions.IsProduction` does not have to be `nullable` if you replace `??` with `||` here.

Lombiq.Hosting.Tenants.EnvironmentRobots/Middlewares/EnvironmentRobotsMiddleware.cs
```diff
@@ -0,0 +1,47 @@
...
+            if (!headerValue.Contains("nofollow")) directives.Add("nofollow");
+
+            if (directives.Any())
+            {
+                context.Response.Headers["X-Robots-Tag"] = $"{headerValue}, {string.Join(", ", directives)}";
```
0liver: This will contain a leading `, ` in case `headerValue` is `""`.

You could work around that by initializing `directives` with `headerValue`, and then changing this line to:
```suggestion
                context.Response.Headers["X-Robots-Tag"] = $"{string.Join(", ", directives)}";
```

0liver (changes requested):

Lombiq.Hosting.Tenants.EnvironmentRobots/Extensions/IHostEnvironmentEnvironmentRobotsExtensions.cs
DemeSzabolcs: It has to be `nullable`. Zoltán already mentioned this: 
> in case ASP.NET Core should be told it runs in Production but not actually (like in staging, that should be as close to production as possible).

For example in staging, if it's not nullable and we are using `||` here it would return `true` every time here, because of ,  `hostEnvironment.IsProduction();` even if we use the custom configuration and set it to false.

0liver: Wow! Good catch. Thanks for the explanation.

Psichorex: In this case I resolve this comment.
For me it was wierd to have a lambda method start on the same column as the parameters I saw indented versions of this in some project.

Lombiq.Hosting.Tenants.EnvironmentRobots.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
Psichorex: I think if sufficient now but as it is not my request please resolve it if it is good for you

Psichorex (approved):

Lombiq.Hosting.Tenants.EnvironmentRobots/Extensions/IHostEnvironmentEnvironmentRobotsExtensions.cs
Psichorex: Please approve if you feel that the conversation is resolved.

PR #71, opened 2023-06-12
OSOE-629: Test Features Guard with recipe import too

### LINKED ISSUE #61: Test Features Guard with recipe import too (OSOE-629)
During UI testing, also test that enabling forbidden features from recipes indeed doesn't enable them, and that conditionally enabled features are enabled and disabled properly with feature states changing via recipe imports too.

This can be done via the JSON import feature from the admin.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-629)

Piedone:
[OSOE-629](https://lombiq.atlassian.net/browse/OSOE-629)
Fixes #61

[OSOE-629]: https://lombiq.atlassian.net/browse/OSOE-629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Hosting.Tenants.FeaturesGuard.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
+++ b/Lombiq.Hosting.Tenants.FeaturesGuard.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
@@ -47,7 +47,7 @@ public static class TestCaseUITestContextExtensions
         // Doing the same checks but with recipes. Starting with a new tenant to make sure that the starting state is
         // correct.
         context.SwitchCurrentTenantToDefault();
-        await SetUpNewTenantAndGoToFeaturesListAsync(context, setupRecipeId, "TestTenant2", "testtenant2");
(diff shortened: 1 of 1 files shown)


Psichorex:
I think we should also include a test which is retrieving the Dictionary from the `appsettings.json` and simultaneously from the `Program.cs` because currently the json part is not tested as per getting it from the appsettings directly. 
But that one will only work as desired after we merge the latest changes into Hosting Tenants with the OC 1.6 upgrade branch so OSOE-548 in PR https://github.com/Lombiq/Hosting-Tenants/pull/55
What dou think?

Piedone:
Both of these configurations are really the same thing: when you add a config to appsettings, it'll be loaded into the same `IShellConfiguration` that you directly interact with from `Program`. I don't really see a point testing that (while it's theoretically possibly that something break in this, that would mean that something is fundamentally broken in ASP.NET Core or OC, what we'd notice by all UI tests, since they all rely on such config working, breaking).

Psichorex (approved):

PR #79, opened 2023-07-20
OSOE-658: Rename MaximumSpace to MaximumSpaceBytes

### LINKED ISSUE #77: Rename MaximumSpace to MaximumSpaceBytes (OSOE-658)
The config `MaximumSpace` doesn't convey what unit it uses. Commonly, we add a suffix for this, so let's do that now as well, and rename it to `MaximumSpaceBytes`.

- [x] Rename all instances of this identifier. This includes not just the configuration key but all internal variables, and methods as well.
- [x] For backward compatibility, the logic should be that first the new key is read, and if there's no value for it, then the old one (and if there's no value for that either, then use the default as we do now).

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-658)

DemeSzabolcs:
[OSOE-658](https://lombiq.atlassian.net/browse/OSOE-658)
Fixes #77

[OSOE-658]: https://lombiq.atlassian.net/browse/OSOE-658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Hosting.Tenants.MediaStorageManagement/Startup.cs
+++ b/Lombiq.Hosting.Tenants.MediaStorageManagement/Startup.cs
@@ -20,7 +20,7 @@ public class Startup : StartupBase
     {
         var maximumStorageQuotaBytes =
             _shellConfiguration.GetValue<long?>(
-                "Lombiq_Hosting_Tenants_MediaStorageManagement:Media_Storage_Management_Options:MaximumSpaceBytes") ??
+                "Lombiq_Hosting_Tenants_MediaStorageManagement:MaximumStorageQuotaBytes") ??
             _shellConfiguration.GetValue<long?>(
                 "Lombiq_Hosting_Tenants_MediaStorageManagement:Media_Storage_Management_Options:MaximumSpace");
         services.Configure<MediaStorageManagementOptions>(options =>
--- a/Lombiq.Hosting.Tenants.MediaStorageManagement/Startup.cs
+++ b/Lombiq.Hosting.Tenants.MediaStorageManagement/Startup.cs
@@ -18,11 +18,13 @@ public class Startup : StartupBase
 
     public override void ConfigureServices(IServiceCollection services)
     {
-        var maximumStorageQuota =
+        var maximumStorageQuotaBytes =
+            _shellConfiguration.GetValue<long?>(
+                "Lombiq_Hosting_Tenants_MediaStorageManagement:Media_Storage_Management_Options:MaximumSpaceBytes") ??
             _shellConfiguration.GetValue<long?>(
                 "Lombiq_Hosting_Tenants_MediaStorageManagement:Media_Storage_Management_Options:MaximumSpace");
         services.Configure<MediaStorageManagementOptions>(options =>
-            options.MaximumStorageQuota = maximumStorageQuota ?? MaximumStorageQuota);
+            options.MaximumStorageQuotaBytes = maximumStorageQuotaBytes ?? MaximumStorageQuotaBytes);
 
         services.AddScoped<IMediaStorageQuotaService, MediaStorageQuotaService>();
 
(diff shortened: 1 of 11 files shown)


Lombiq.Hosting.Tenants.MediaStorageManagement/Startup.cs
```diff
@@ -18,11 +18,13 @@ public class Startup : StartupBase
...
     {
-        var maximumStorageQuota =
+        var maximumStorageQuotaBytes =
+            _shellConfiguration.GetValue<long?>(
+                "Lombiq_Hosting_Tenants_MediaStorageManagement:Media_Storage_Management_Options:MaximumSpaceBytes") ??
             _shellConfiguration.GetValue<long?>(
                 "Lombiq_Hosting_Tenants_MediaStorageManagement:Media_Storage_Management_Options:MaximumSpace");
         services.Configure<MediaStorageManagementOptions>(options =>
-            options.MaximumStorageQuota = maximumStorageQuota ?? MaximumStorageQuota);
+            options.MaximumStorageQuotaBytes = maximumStorageQuotaBytes ?? MaximumStorageQuotaBytes);
```
Piedone, line 27: I just noticed that the option is actually called StorageQuota, while the config key Size... That's bad. Please fix this by renaming the latter to use the StorageQuota terminology (and all "space" identifiers too).

Furthermore, the key has wrong underscores, so it should be `"Lombiq_Hosting_Tenants_MediaStorageManagement:MediaStorageManagementOptions:MaximumStorageQuotaBytes"`. But! We don't actually need this indirection since we won't have a lot of layers for config here, so rather name it `"Lombiq_Hosting_Tenants_MediaStorageManagement:MaximumStorageQuotaBytes"`.

Lombiq.Hosting.Tenants.MediaStorageManagement/Views/UploadFileSize.cshtml
```diff
@@ -21,7 +21,7 @@ $(function () {
             return;
         }
         model.errorMessage =
-        @T["You may only store {0} MB of Media files for your site, and it seems that you’d just go over that. But don’t worry! If you delete some large files, you should be able to upload new ones. Also, you can contact us for a larger quota.", Model.MaximumSpace].Json();
+        @T["You may only store {0} MB of Media files for your site, and it seems that you’d just go over that. But don’t worry! If you delete some large files, you should be able to upload new ones. Also, you can contact us for a larger quota.", Model.MaximumSpaceBytes].Json();
```
Piedone: The text is about MB while the value is B. Let's use actual MBs.

Piedone (changes requested):

Lombiq.Hosting.Tenants.MediaStorageManagement/Startup.cs
DemeSzabolcs: > while the config key Size

Where is this config key? I can't find it.

> But! We don't actually need this indirection since we won't have a lot of layers for config here, so rather name it "Lombiq_Hosting_Tenants_MediaStorageManagement:MaximumStorageQuotaBytes".

Then I need to delete `MediaStorageManagementOptions.cs` no?

Lombiq.Hosting.Tenants.MediaStorageManagement/Views/UploadFileSize.cshtml
DemeSzabolcs: The value is MB too, just the variable name is wrong.

Lombiq.Hosting.Tenants.MediaStorageManagement/Startup.cs
Piedone: I mean, the property in `MediaStorageManagementOptions` uses different terminology than the configuration key read above. I'm asking you to please rename the configuration key to `"Lombiq_Hosting_Tenants_MediaStorageManagement:MaximumStorageQuotaBytes"` (so it matches what's in the options class), and adapt the terminology used by the other identifiers (variables, properties) too.

DemeSzabolcs: Ah I guess you wanted to write Space but you wrote

> Size

That's why I was confused.

Piedone (approved):

PR #95, opened 2023-10-19
NEST-486: Renaming EmailQuotaIndex columns and dropping index

wAsnk:
[NEST-486](https://lombiq.atlassian.net/browse/NEST-486)

[NEST-486]: https://lombiq.atlassian.net/browse/NEST-486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Hosting.Tenants.EmailQuotaManagement/Migrations/EmailQuotaMigrations.cs
```diff
@@ -25,4 +14,30 @@ public int UpdateFrom1()
 
         return 2;
     }
+
+    public int UpdateFrom2()
+    {
+        // Renaming for safety reasons.
```
Piedone: What are the safety reasons? Why can't we drop the table regardless of its column names?

Lombiq.Hosting.Tenants.EmailQuotaManagement/Migrations/EmailQuotaMigrations.cs
```diff
@@ -1,22 +1,11 @@
...
-using YesSql.Sql;
+using System.Data.Common;
 
 namespace Lombiq.Hosting.Tenants.EmailQuotaManagement.Migrations;
 
 public class EmailQuotaMigrations : DataMigration
 {
-    public int Create()
-    {
-        SchemaBuilder.CreateMapIndexTable<EmailQuotaIndex>(
```
Piedone, line 12: Delete `EmailQuotaIndex` too.

Lombiq.Hosting.Tenants.EmailQuotaManagement/Migrations/EmailQuotaMigrations.cs
```diff
@@ -25,4 +14,30 @@ public int UpdateFrom1()
 
         return 2;
     }
+
+    public int UpdateFrom2()
+    {
+        // Renaming for safety reasons.
+        TryRenameColumn("LastReminder", nameof(EmailQuotaIndex.LastReminderUtc));
```
Piedone: You can use string names in this class now so `EmailQuotaIndex` can be deleted.

Piedone (changes requested):

Lombiq.Hosting.Tenants.EmailQuotaManagement/Migrations/EmailQuotaMigrations.cs
wAsnk: What would happen if I don't rename the column names and then we are looking for the data in the Documents table, but it still has the old column names saved as JSON?

Piedone: The Document table doesn't reference column names from the index but property names from the `EmailQuota` model.

wAsnk: Yes, but the EmailQuota model property names were changed as well.

wAsnk: Hmm, but those might be already saved as the new name I guess. Nvm then.

Piedone: The index and the document might use the same names, but these are still different identifiers.

Lombiq.Hosting.Tenants.EmailQuotaManagement/Migrations/EmailQuotaMigrations.cs
```diff
@@ -1,28 +1,23 @@
-using Lombiq.Hosting.Tenants.EmailQuotaManagement.Indexes;
 using OrchardCore.Data.Migration;
-using System;
-using YesSql.Sql;
 
 namespace Lombiq.Hosting.Tenants.EmailQuotaManagement.Migrations;
 
 public class EmailQuotaMigrations : DataMigration
 {
-    public int Create()
```
Piedone, line 10: `Create()` needs to remain for new installations but empty and return 3.

Piedone (approved):

PR #102, opened 2023-11-23
OSOE-708: Email Quota Management UX improvements in Lombiq.Hosting.Tenants

### LINKED ISSUE #94: Email Quota Management UX improvements (OSOE-708)
- Change [this wording](https://github.com/Lombiq/Hosting-Tenants/blob/819285d015c422c576e8d9aad6a992f536047dfe/Lombiq.Hosting.Tenants.EmailQuotaManagement/Views/EmailSettingsQuotaMessage.cshtml#L3) to "You've sent {0} emails from the total of {1} **this month**."
- Display [DashboardQuotaMessage](https://github.com/Lombiq/Hosting-Tenants/blob/819285d015c422c576e8d9aad6a992f536047dfe/Lombiq.Hosting.Tenants.EmailQuotaManagement/Views/DashboardQuotaMessage.cshtml#L6) permanently at the top of each admin page, not just /admin.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-708)

wAsnk:
[OSOE-708](https://lombiq.atlassian.net/browse/OSOE-708)
Fixes #94

[OSOE-708]: https://lombiq.atlassian.net/browse/OSOE-708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Hosting.Tenants.EmailQuotaManagement.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
+++ b/Lombiq.Hosting.Tenants.EmailQuotaManagement.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
@@ -48,10 +48,14 @@ public static class TestCaseUITestContextExtensions
             else if (warningLevel >= 80)
             {
                 await context.GoToDashboardAsync();
-                context.CheckExistence(
-                    By.XPath($"//p[contains(@class,'alert-warning')]" +
-                        $"[contains(.,'It seems that your site sent out {warningLevel.ToTechnicalString()}% of e-mail')]"),
-                    exists: true);
+                CheckMessageExistence(context, warningLevel.ToTechnicalString());
+
+                await context.GoToContentItemsPageAsync();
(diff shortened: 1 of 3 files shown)


Psichorex (approved):

Psichorex:
My only question here is that the original feature was about to display a message about the quota on the dashboard. This feature was extended to show up not only on the dashboard but all the admin pages. I think the UITest should cover that because it is only checking on the dashboard which is the original feature but not the extended. Otherwise approved.

wAsnk:
> My only question here is that the original feature was about to display a message about the quota on the dashboard. This feature was extended to show up not only on the dashboard but all the admin pages. I think the UITest should cover that because it is only checking on the dashboard which is the original feature but not the extended. Otherwise approved.

Yeah you are right, I'll adjust it with some additional admin pages to check.

PR #106, opened 2024-01-18
OSOE-751: Upgrade to Orchard Core 1.8

Psichorex:
[OSOE-751](https://lombiq.atlassian.net/browse/OSOE-751)

[OSOE-751]: https://lombiq.atlassian.net/browse/OSOE-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Hosting.Tenants.EmailQuotaManagement/Services/EmailQuotaService.cs
+++ b/Lombiq.Hosting.Tenants.EmailQuotaManagement/Services/EmailQuotaService.cs
@@ -1,4 +1,4 @@
-﻿using Lombiq.Hosting.Tenants.EmailQuotaManagement.Models;
+using Lombiq.Hosting.Tenants.EmailQuotaManagement.Models;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Options;
@@ -91,22 +91,22 @@ public class EmailQuotaService : IEmailQuotaService
             // Need to set default value otherwise the database might complain about being 01/01/0001 out of range.
             LastReminderUtc = _clock.UtcNow.AddMonths(-1),
         };
-        _session.Save(currentQuota);
+        await _session.SaveAsync(currentQuota);
 
         return currentQuota;
     }
 
-    public void IncreaseEmailUsage(EmailQuota emailQuota)
+    public Task IncreaseEmailUsageAsync(EmailQuota emailQuota)
     {
         emailQuota.CurrentEmailUsageCount++;
-        _session.Save(emailQuota);
+        return _session.SaveAsync(emailQuota);
     }
 
-    public void SetQuotaOnEmailReminder(EmailQuota emailQuota)
+    public Task SetQuotaOnEmailReminderAsync(EmailQuota emailQuota)
     {
         emailQuota.LastReminderUtc = _clock.UtcNow;
         emailQuota.LastReminderPercentage = emailQuota.CurrentUsagePercentage(GetEmailQuotaPerMonth());
-        _session.Save(emailQuota);
+        return _session.SaveAsync(emailQuota);
     }
 
     public bool ShouldSendReminderEmail(EmailQuota emailQuota, int currentUsagePercentage)
@@ -128,10 +128,10 @@ public class EmailQuotaService : IEmailQuotaService
             emailQuota.LastReminderPercentage >= 100);
     }
 
-    public void ResetQuota(EmailQuota emailQuota)
+    public Task ResetQuotaAsync(EmailQuota emailQuota)
     {
         emailQuota.CurrentEmailUsageCount = 0;
-        _session.Save(emailQuota);
+        return _session.SaveAsync(emailQuota);
     }
 
     public int GetEmailQuotaPerMonth() =>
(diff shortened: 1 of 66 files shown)


Lombiq.Hosting.Tenants.EmailQuotaManagement/Services/EmailQuotaService.cs
```diff
@@ -19,54 +19,38 @@
...
-        _userManager = userManager;
-    }
+    private readonly EmailQuotaOptions _emailQuotaOptions = emailQuotaOptions.Value;
+    private readonly SmtpSettings _smtpOptions = smtpOptions.Value;
 
     public async Task<IEnumerable<string>> GetUserEmailsForEmailReminderAsync()
     {
         // Get users with site owner permission.
-        var roles = await _roleService.GetRolesAsync();
+        var roles = await roleService.GetRolesAsync();
```
Piedone: Keep using underscored named for private fields everywhere.

Lombiq.Hosting.Tenants.Management/Controllers/ShellSettingsEditorController.cs
```diff
@@ -119,8 +102,8 @@ public async Task<IActionResult> Edit(ShellSettingsEditorViewModel model)
         // not save settings that has a key with multiple sections, see
         // https://github.com/OrchardCMS/OrchardCore/issues/14481. Once this is fixed, we can get rid of the locking and
```
Piedone: Don't forget to remove the workaround, since this issue is now fixed.

Lombiq.Hosting.Tenants.EmailQuotaManagement/Services/EmailQuotaService.cs
Psichorex: That would have an affair with https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1313.md.
Also it would not be compatible with VS built in convert to primary constructor feature.
Also this would mean to change like 3000 private field back to _ version because VS removed them.
We only needed _ version for the sheer reason to differentiate between the injection and the private field.
So because of these I would not recommend switching back to that. Also they look wierd in parameters.

Lombiq.Hosting.Tenants.Management/Controllers/ShellSettingsEditorController.cs
Psichorex: https://github.com/OrchardCMS/OrchardCore/issues/15184

I didn't find a way to make this actually work. Also it broke after 1.8. The problem seems to be different from the described in the issue.

My problem was that when we wanted to delete keys because they are no longer in the editor I couldn't save them.
No matter what kind of `Update()` or `Save` method I called. I could only delete after referencing `ShellConfiguration[key]` directly instead of using a `Dictionary`. Tested in the Source aswell.

Lombiq.Hosting.Tenants.EmailQuotaManagement/Services/EmailQuotaService.cs
Piedone: Please bring this up with the StyleCop team, in an issue if a similar one doesn't exist already. We definitely don't want to change the private field naming convention partially like this (and thus have it mixed), that's out of the question. If StyleCop will be able to support exceptions to this rule for primary constructors then great, but if not, then we shouldn't use primary constructors (having to add suppressions for each of them is a no-go too).

Lombiq.Hosting.Tenants.Management/Controllers/ShellSettingsEditorController.cs
Piedone: Document https://github.com/OrchardCMS/OrchardCore/issues/15184 here instead, then, expanding on the current docs, so during a later upgrade we can understand what needs to be changed.

Psichorex: I added a comment in our code with link to the issue and followed up there.

Lombiq.Hosting.Tenants.EmailQuotaManagement/Services/EmailQuotaService.cs
Psichorex: I resolve this thus this is not a question anymore.

Lombiq.Hosting.Tenants.Management/Controllers/ShellSettingsEditorController.cs
Psichorex: I removed the workaround after PPing with Krisztián he confirmed that the way it is now should have been the original way.

Lombiq.Hosting.Tenants.Management/Filters/ForbiddenTenantsFilter.cs
```diff
@@ -50,11 +51,9 @@ public Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecuti
             var requestUrlHost = context.HttpContext.Request.Form[nameof(CreateApiViewModel.RequestUrlHost)].ToString();
             var hosts = requestUrlHost.Split(',').Select(host => host.Trim());
 
-            List<string> unacceptableHostnames = new();
+            List<string> unacceptableHostnames = [.. hosts.Where(hostname => forbiddenRequestUrlHosts.Contains(hostname))];
```
Piedone:
```suggestion
            var unacceptableHostnames = hosts.Where(hostname => forbiddenRequestUrlHosts.Contains(hostname)).ToList();
```

Piedone (changes requested):

Piedone (approved):

Piedone (approved):

Piedone (approved):

Piedone (approved):

PR #156, opened 2025-04-15
NEST-561: Adding staggered maintenance

wAsnk:
[NEST-561](https://lombiq.atlassian.net/browse/NEST-561)

[NEST-561]: https://lombiq.atlassian.net/browse/NEST-561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.Hosting.Tenants.Maintenance/Assets/Scripts/StaggeredMaintenance.js
```diff
@@ -0,0 +1,151 @@
+/* global bootstrap */
```
DemeSzabolcs: I think you can delete this.

Lombiq.Hosting.Tenants.Maintenance/Assets/Scripts/StaggeredMaintenance.js
```diff
@@ -0,0 +1,151 @@
...
+
+    if (countdownTimer !== null) {
+        clearInterval(countdownTimer);
+        countdownTimer = null;
+    }
+
+    updateCountdownDisplay(true);
+}
+
+// === Animation ===
```
DemeSzabolcs: Edit the comments so they follow the usual `// Comment.` format. So dot at the end.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
```diff
@@ -0,0 +1,84 @@
...
+                .WithSettings(new NumericFieldSettings
+                {
+                    Hint = "Calculated from processed tenants count and all tenant count.",
+                }))
+            .WithField<NumericField>(nameof(StaggeredMaintenancePart.AllTenantCount), field => field
+                .WithDisplayName("All Tenant Count")
+                .WithSettings(new NumericFieldSettings
+                {
+                    Hint = "The number of tenants to be processed. This is the total number of running tenants in the system.",
+                }))
```
DemeSzabolcs: Are these supposed to be editable? Because they are, so if not, it would be good to hide those on the edit page. Because if I understand correctly, these are calculated.

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/Index.cshtml
```diff
@@ -0,0 +1,39 @@
...
+        typeof(AdminController).ControllerName(),
+        new { area = FeatureNames.Maintenance })">
+    @await Html.PartialAsync("StaggeredMaintenanceDetails", Model)
+</div>
+
+<style at="Head">
+    #maintenance-progress-bar {
+        transition: width 1s ease;
+    }
+</style>
```
DemeSzabolcs: Yeah, if this is the only styling here, it's not worth adding a CSS for this.

Lombiq.Hosting.Tenants.Maintenance/Models/StaggeredMaintenanceOptions.cs
```diff
@@ -0,0 +1,8 @@
+namespace Lombiq.Hosting.Tenants.Maintenance.Models;
+
+public class StaggeredMaintenanceOptions
+{
+    public int? TimeSpanMilliseconds { get; set; }
```
DemeSzabolcs: Shouldn't this be named to `TimeSpanBetweenBatches`? Or `TimeSpanMillisecondsBetweenBatches`  to be accurate.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
```diff
@@ -0,0 +1,84 @@
...
+                .WithDisplayName("Time Span Between Batches")
+                .WithSettings(new TimeFieldSettings
+                {
+                    // 1 second.
+                    Step = "1",
+                    Hint = "The time span between batches of tenants to be processed in hh:mm:ss format. " +
+                           "If StaggeredMaintenanceOptions is set, this value will be ignored.",
+                }))
+            .WithField<BooleanField>(nameof(StaggeredMaintenancePart.RunParallel), field => field
+                .WithSettings(new BooleanFieldSettings
```
DemeSzabolcs: Add `.WithDisplayName("Run Parallel")`

DemeSzabolcs (changes requested):

Lombiq.Hosting.Tenants.Maintenance/Assets/Scripts/StaggeredMaintenance.js
wAsnk: I'll get "D:\Projects\DotNest-Tenants-Core-3\src\Modules\Lombiq.Hosting.Tenants\Lombiq.Hosting.Tenants.Maintenance\Assets\Scripts\StaggeredMaintenance.js(77,81): warning no-undef: 'bootstrap' is not defined."
lint error otherwise.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
wAsnk: I thought about it too, but it doesn't really do much apart from displaying and displaying logic. And taking into consideration that this is a maintenance it will be handled by someone who knows what they are doing. So I think it's fine leaving it like this, also it gives better configureablity.

DemeSzabolcs (approved):

Lombiq.Hosting.Tenants.Maintenance/Services/IStaggeredMaintenanceService.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// default tenant. Starts the maintenance process for all tenants in a staggered manner, with a delay between each
+    /// batch of tenants.
+    /// </summary>
+    Task<StaggeredMaintenancePart> RunScheduledMaintenanceForAllTenantAsync(bool newVersion = false, bool reset = false);
+
+    /// <summary>
+    /// Gets or creates the staggered maintenance content item. This is done to avoid creating the staggered maintenance
+    /// content item multiple times. We are using a content item structure so we can easily have an editor for it.
+    /// </summary>
+    Task<ContentItem> GetorCreateStaggeredMaintenanceAsync();
```
BenedekFarkas:
```suggestion
    Task<ContentItem> GetOrCreateStaggeredMaintenanceAsync();
```

Lombiq.Hosting.Tenants.Maintenance/Handlers/StaggeredMaintenanceHandler.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        var staggeredMaintenancePart = contentItem.As<StaggeredMaintenancePart>();
+        if (staggeredMaintenancePart == null) return Task.CompletedTask;
+
+        if (staggeredMaintenancePart.IsRunning())
+        {
+            context.Updater.ModelState.AddModelError(
+                nameof(StaggeredMaintenancePart.IsRunning),
+                T["You can't save this until a staggered maintenance is running."]);
```
BenedekFarkas:
```suggestion
                T["You can't save this while a staggered maintenance is running."]);
```

Lombiq.Hosting.Tenants.Maintenance/Models/StaggeredMaintenanceOptions.cs
```diff
@@ -0,0 +1,8 @@
+namespace Lombiq.Hosting.Tenants.Maintenance.Models;
+
+public class StaggeredMaintenanceOptions
+{
+    public int? MillisecondBetweenBatches { get; set; }
```
BenedekFarkas: This can be just seconds for simplicity.
```suggestion
    public int? BatchIntervalSeconds { get; set; }
```

Lombiq.Hosting.Tenants.Maintenance/Models/StaggeredMaintenancePart.cs
```diff
@@ -0,0 +1,85 @@
...
+using OrchardCore.Modules;
+using System;
+using System.Collections.Generic;
+
+namespace Lombiq.Hosting.Tenants.Maintenance.Models;
+
+public class StaggeredMaintenancePart : ContentPart
+{
+    public NumericField ProcessingStep { get; } = new() { Value = 1 };
+    public TimeField TimeSpanBetweenBatches { get; set; } = new() { Value = TimeSpan.FromSeconds(0) };
```
BenedekFarkas:
```suggestion
    public TimeField BatchInterval { get; set; } = new() { Value = TimeSpan.FromSeconds(0) };
```

Lombiq.Hosting.Tenants.Maintenance/Models/StaggeredMaintenancePart.cs
```diff
@@ -0,0 +1,85 @@
...
+using OrchardCore.ContentManagement;
+using OrchardCore.Modules;
+using System;
+using System.Collections.Generic;
+
+namespace Lombiq.Hosting.Tenants.Maintenance.Models;
+
+public class StaggeredMaintenancePart : ContentPart
+{
+    public NumericField ProcessingStep { get; } = new() { Value = 1 };
```
BenedekFarkas: Isn't this the batch size?

Lombiq.Hosting.Tenants.Maintenance/Readme.md
```diff
@@ -196,3 +196,33 @@ The following configuration options are available to set the enabled features:
   }
 }
 ```
+
+### `Lombiq.Hosting.Tenants.Maintenance.StaggeredMaintenance`
+
+Adds a page to the admin under Multi-Tenancy/Maintenance that allows you to start all tenants shells in a staggered way. This is useful when you have a lot of tenants and you want to start them in a staggered way to avoid overwhelming the database or other resources. Starting up all the tenants ensures that they get updated to the newest migration steps and also all the above maintenance tasks run.
+
+You can edit the maintenance options directly on the Multi-Tenancy/Maintenance page if you click on the edit button. You can set the batches that the tenants should be run in and the waiting time span between the batches. You can set if you want to run it in parallel or not. If you set it to parallel it will start the Processing Step amount of tenants at once, otherwise it will start tenants one-by-one and then wait the time span after the Processing Step amount of tenants has been started.
```
BenedekFarkas: Please simplify this with a bullet list to list the options and a brief explanation for each.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
```diff
@@ -0,0 +1,85 @@
...
+                .WithDisplayName("Time Span Between Batches")
+                .WithSettings(new TimeFieldSettings
+                {
+                    // 1 second.
+                    Step = "1",
+                    Hint = "The time span between batches of tenants to be processed in hh:mm:ss format. " +
+                           "If StaggeredMaintenanceOptions is set, this value will be ignored.",
+                }))
+            .WithField<BooleanField>(nameof(StaggeredMaintenancePart.RunParallel), field => field
+                .WithDisplayName("Run Parallel")
```
BenedekFarkas:
```suggestion
                .WithDisplayName("Run In Parallel")
```
BTW currently (while it's set to "Run Parallel") it's displayed as "RunParallel".

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/StaggeredMaintenanceDetails.cshtml
```diff
@@ -0,0 +1,167 @@
...
+@{
+    var part = Model.ContentItem.As<StaggeredMaintenancePart>();
+    var currentPercentage = part.ProgressPercentage.Value;
+
+}
+
+<div class="buttons mt-4">
+    @await DisplayAsync(await New.StaggeredMaintenance__NewVersionButton(Part: part))
+    @await DisplayAsync(await New.StaggeredMaintenance__ContinueButton(Part: part))
+    @await DisplayAsync(await New.StaggeredMaintenance__CancelButton(Part: part))
```
BenedekFarkas: Are Cancel and Continue the opposites of each other? In that case Cancel should be Pause instead and only display the one that you could actually click.

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/Index.cshtml
```diff
@@ -0,0 +1,39 @@
+﻿@using Lombiq.Hosting.Tenants.Maintenance.Constants
+@using Lombiq.Hosting.Tenants.Maintenance.Controllers
+@using OrchardCore.Mvc.Core.Utilities
+@model OrchardCore.ContentManagement.ContentItem
+
+<zone Name="Title"><h1>@RenderTitleSegments(T["Staggered Maintenance"])</h1></zone>
+<div class="d-flex align-items-center gap-3 mb-3">
+    <div class="form-check form-switch">
+        <input class="form-check-input" type="checkbox" id="autoRefreshToggle">
+        <label class="form-check-label" for="autoRefreshToggle">@T["Auto Refresh"]</label>
```
BenedekFarkas: I didn't think about adding async refresh at all, but it's pretty neat that you added it!

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
BenedekFarkas, line 1: These fields should not be editable, only their values displayed, because they are computed: Progress Percentage, All Tenant Count, Canceled, Run In Parallel, Started, Finished.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
```diff
@@ -0,0 +1,85 @@
...
+            .WithField<NumericField>(nameof(StaggeredMaintenancePart.ProcessingStep), field => field
+                .WithDisplayName("Processing Step")
+                .WithSettings(new NumericFieldSettings
+                {
+                    Hint = "The number of tenants to be processed in each step or in case of parallel processing the" +
+                           " number of tenants started at once.",
+                }))
+            .WithField<TimeField>(nameof(StaggeredMaintenancePart.TimeSpanBetweenBatches), field => field
+                .WithDisplayName("Time Span Between Batches")
+                .WithSettings(new TimeFieldSettings
```
BenedekFarkas: It should just be a simple counter to define the number of seconds (also see my other comment amount not needing millisecond-precision). BTW for millisecond-precision this editor wouldn't be suitable anyway.

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/StaggeredMaintenanceDetails.cshtml
```diff
@@ -0,0 +1,167 @@
...
+
+            <dt class="col-sm-6">@T["Processed tenants:"]</dt>
+            <dd class="col-sm-6">@part.ProcessedTenantIds.Count</dd>
+        </dl>
+    </div>
+</div>
+
+<div class="accordion mt-4" id="maintenanceAccordion">
+    <!-- Errors Accordion Item -->
+    <div class="accordion-item">
```
BenedekFarkas: Opening an accordion turns its header blue and then it cannot be closed.

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/StaggeredMaintenanceDetails.cshtml
```diff
@@ -0,0 +1,167 @@
...
+                else
+                {
+                    <p class="text-muted">@T["No errors found. There might be errors caught, logged and not thrown again. Please check the logs."]</p>
+                }
+            </div>
+        </div>
+    </div>
+
+    <!-- Versions Accordion Item -->
+    <div class="accordion-item">
```
BenedekFarkas: Can we list them alphabetically or is there a reason for not doing so? I assume they are in execution order currently.

Lombiq.Hosting.Tenants.Maintenance/AdminMenu.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    private readonly ShellSettings _shellSettings;
+
+    public AdminMenu(IHttpContextAccessor hca, IStringLocalizer<AdminMenu> stringLocalizer, ShellSettings shellSettings)
+        : base(hca, stringLocalizer) => _shellSettings = shellSettings;
+
+    protected override void Build(NavigationBuilder builder)
+    {
+        // Don't add the menu item on non-default tenants.
+        if (!_shellSettings.IsDefaultShell())
```
BenedekFarkas, line 20: Instead of hiding the menu item, this feature should only be allowed on the default tenant. Currently if you try to navigate to `/Admin/StaggeredMaintenance/Index` on an other tenant, you'll get an exception.

BenedekFarkas (changes requested):

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
wAsnk: Szabolcs had the same idea:
https://github.com/Lombiq/Hosting-Tenants/pull/156#discussion_r2048878583

> I thought about it too, but it doesn't really do much apart from displaying and displaying logic. And taking into consideration that this is a maintenance it will be handled by someone who knows what they are doing. So I think it's fine leaving it like this, also it gives better configurability.

But as two of you think the same, I'll do it. 😄

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/StaggeredMaintenanceDetails.cshtml
wAsnk: Not sure why, I'm using the same template as it is on the bootstrap docs. I think it might be related to the version OC uses. I wouldn't spend time on it.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
wAsnk: You can't define milliseconds, because the step in the TimeFieldSettings is set to "1".

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/StaggeredMaintenanceDetails.cshtml
wAsnk: I would leave them both as it might be useful to have them if something unexpected happens. Then you simply remove the disabled attribute. As this is something only we would use I think it's fine like this.

BenedekFarkas: OK, leave them, but "Cancel" should still be "Pause", since it doesn't actually cancel the process.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
BenedekFarkas: BTW taking this into consideration, making the configuration a content item seems a bit overkill.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
BenedekFarkas: I know, that's why I'm saying that it's not suitable for milliseconds. Previously this configuration was stored as `StaggeredMaintenanceOptions.MillisecondBetweenBatches` until I suggested to rename it to `BatchIntervalSeconds` (and change the logic accordingly).

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/StaggeredMaintenanceDetails.cshtml
BenedekFarkas: If you're using a certain version of a library (wherever it comes from), then you need to write code that's compatible with that version.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
wAsnk: But it's nice to edit it like this, isn't it? 😄

BenedekFarkas: In theory yes, but you have a custom type (of which you'll only ever have one item) with a bunch of content fields, most of which are just displayed, never edited. Site settings is more appropriate here (it can be a different tab on the same screen).

wAsnk: I tried site settings, but they behave differently. It won't update for display until the maintenance is not fully finished. That's why I decided to use a content item directly.

BenedekFarkas: OK, let's do content item then, but then you can remove most of the content fields (which shouldn't be user-editable) and just use properties to store those values instead.

wAsnk: Okay

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/Index.cshtml
```diff
@@ -23,11 +29,11 @@
...
 <div id="maintenance-status-container"
-    data-maintenance-url="@Url.Action(
-        nameof(AdminController.GetPartialView),
-        typeof(AdminController).ControllerName(),
-        new { area = FeatureNames.Maintenance })">
-    @await Html.PartialAsync("StaggeredMaintenanceDetails", Model)
+     data-maintenance-url="@Url.Action(
+                               nameof(AdminController.GetPartialView),
+                               typeof(AdminController).ControllerName(),
+                               new { area = FeatureNames.Maintenance })">
```
DemeSzabolcs: Code styling here is wrong.

DemeSzabolcs (changes requested):

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/Index.cshtml
wAsnk: Sometimes I hate Rider auto styling :(

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/Index.cshtml
BenedekFarkas: @wAsnk with a fresh setup I tested DotNest Tenants (which locally only has 3 tenants set up), enabled auto-refresh and set the refresh interval to 5 seconds. It runs successfully, but it doesn't refresh anything. I suspect that it might be because it finishes so quickly, so maybe the need to refresh is determined by the percentage (which doesn't change in this case, but other properties displayed should)?

Lombiq.Hosting.Tenants.Maintenance/AdminMenu.cs
BenedekFarkas: `Admin/StaggeredTenantWakeUp/Index` is now accessible on tenants other than Default, it should only be allowed to run (be enabled) on the Default tenant.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    public async Task<int> CreateAsync()
+    {
+        await _contentDefinitionManager.AlterPartDefinitionAsync(nameof(StaggeredTenantWakeUpPart), part => part
+            .WithField<NumericField>(nameof(StaggeredTenantWakeUpPart.BatchSize), field => field
+                .WithDisplayName("Batch Size")
+                .WithSettings(new NumericFieldSettings
+                {
+                    Hint = "The number of tenants to be processed in each step or in case of parallel processing the" +
+                           " number of tenants started at once. If StaggeredTenantWakeUpOptions is set, these values will be ignored.",
```
BenedekFarkas: I just noticed this:
`If StaggeredTenantWakeUpOptions is set, these values will be ignored.`

Is this really the case? That's not how the settings UI usually works: If you have appsettings, then that's taken into account without having to go to the UI, but the values set in the UI should take higher precedence (and it should be initialized to appsettings, if exists), otherwise it's useless if you have anything set in appsettings.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
```diff
@@ -0,0 +1,57 @@
...
+                .WithSettings(new BooleanFieldSettings
+                {
+                    Hint = "Indicates whether the staggered tenant wake-up process should run in parallel or not.",
+                }))
+            .WithField<NumericField>(nameof(StaggeredTenantWakeUpPart.CurrentVersion), field => field
+                .WithDisplayName("Current Version")
+                .WithSettings(new NumericFieldSettings
+                {
+                    Hint = "The current version of the staggered tenant wake-up process.",
+                }))
```
BenedekFarkas: Is there a use-case for allowing this to be edited? Similarly, "Reset current version" seems unnecessary to me, because the version number is just a counter, but the code/logic that runs is independent of it (unless you re-deploy the app with different logic, of course).

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
BenedekFarkas: This is still a TimeField, but IMO it's confusing UI-wise - this is how it looks like when set to 20 seconds:
image
I still think that a simple (unsigned) integer value, i.e. NumericField (BatchIntervalSeconds) with a value at least 0 would be a lot more straightforward.

BenedekFarkas (changes requested):

Lombiq.Hosting.Tenants.Maintenance/Readme.md
```diff
@@ -196,3 +196,41 @@ The following configuration options are available to set the enabled features:
...
+
+```json
+{
+  "OrchardCore": {
+    "Lombiq_Hosting_Tenants_Maintenance": {
+      "StaggeredTenantWakeUp": {
+        "BatchIntervalSeconds": 5,
+        "BatchSize": 5,
+        "RunParallel": false,
+        "RunOnStartup": true
```
BenedekFarkas, line 228: Any reason for not including this in the admin UI editor? UX-wise it's better to have parity between them.

Lombiq.Hosting.Tenants.Maintenance/Readme.md
```diff
@@ -196,3 +196,41 @@ The following configuration options are available to set the enabled features:
...
+      }
+    }
+  }
+}
+```
+
+- RunOnStartup: When set to true, the staggered tenant wake-up process will start a new version if a new deployment was done, or it will continue the current wake-up process if it hasn't finished yet. This is useful if you want to run the staggered tenant wake-up process automatically when the application starts. If you set this to false, you will have to start it manually from the admin page.
```
BenedekFarkas: Based on the text (haven't tested yet): If a new deployment happens, the process should start from scratch, otherwise some tenants might not run the changes that were just deployed.

BenedekFarkas (changes requested):

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
wAsnk: Hmm, okay. That's indeed true, it is better UX to do it this way!

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
wAsnk: I think I removed the editing possibility only for those that you mentioned before. Now deleted this as well.

Yeah it is debatable whether you need the "Reset current version" functionality. I removed it now.

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
wAsnk: This is weird, because I had it working like 00:00:20 for seconds, but if this isn't working like you showed, then it is not worth fighting for this from my side 😄 So I changed it now to a NumericField.

Lombiq.Hosting.Tenants.Maintenance/AdminMenu.cs
wAsnk: Yeah I messed up, I added the feature to be enabled from Program.cs. I thought that DefaultTenantOnly in manifest actually does what I want.

Lombiq.Hosting.Tenants.Maintenance/Readme.md
wAsnk: Added it now

Lombiq.Hosting.Tenants.Maintenance/Readme.md
wAsnk: Yes!

Lombiq.Hosting.Tenants.Maintenance/StaggeredMigrations.cs
BenedekFarkas: It's browser-dependent too (the screenshot is from Brave), but for simplicity's sake IMO seconds is enough. The most that could make sense to configure here is couple minutes (~ few hundred seconds), but in reality (for us, at least) it will be definitely be less than 60 seconds.

Lombiq.Hosting.Tenants.Maintenance/AdminMenu.cs
wAsnk: Ahh, this is a misery. Do we have experience in enabling a feature only for default tenant, but not from a recipe? Because even if I check if the tenant is the default tenant in the startup so the services are not added, the controller will be activated anyway. 

If I delete the activation of this feature from program.cs, then you need to enable the feature manually whenever you do a staging reset, until it is not in prod. Once it is in prod you have to manually enable the feature.

How should we do this?

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/Index.cshtml
wAsnk: The auto refresher checks the percentage, if it is between 0 and 100 and if it is not paused. The problem is when the page refreshes while clicking on Start new, the page loads so fast that within the background job the content item is not updated yet.

I can delete this logic, so it auto refreshes simply on every X seconds without checking a state. (It will still cancel auto refresh if the tab is hidden.)

wAsnk: For now I deleted the logic, so it auto refreshes until it is not disabled or if the tab is not shown.

Lombiq.Hosting.Tenants.Maintenance/AdminMenu.cs
wAsnk: I think the easiest would be to enable the feature manually once on the default tenant.

BenedekFarkas: Sure, that's fine.

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/Index.cshtml
BenedekFarkas: How about refreshing once always after saving page (after the set interval time has passed)?

Lombiq.Hosting.Tenants.Maintenance/AdminMenu.cs
BenedekFarkas: Actually,

> I thought that DefaultTenantOnly in manifest actually does what I want.

should work indeed. Did the controller belong to that feature only? Because it might be that it mostly did what you wanted, but not everything was covered.

wAsnk: Yes, the controller had the Feature attribute as it has it now.

BenedekFarkas: Hmm. Does it work in vanilla OC? Or is it even used there this way (on a Controller)? I have a hunch that since ASP.NET Core has a different DI mechanism (compared to O1), it doesn't work with a controller, but it could be a bug too.

wAsnk: I don't think there is a problem with the feature attribute on the controller. When the feature is not enabled you get 404 like you should. When the feature is enabled, you get the page.

If the feature is enabled through Program.cs (I think that doesn't respect the DefaultTenantOnly property in the manifest), the feature gets enabled on tenants also. But in the Startup of this feature I only let default tenant to add services and other stuff, but for the controller I can't do the same.

So either I check in the controller on each action if it's the default tenant and return 404 if it isn't, then I put back the feature enabling in the Program.cs so it gets enabled on all tenant.
Or we have to enable it manually once in prod.

BenedekFarkas: Ooooh, OK! Manually enabling it for the first time is no problem (make sure that it's in the setup recipe too in DotNest).

Lombiq.Hosting.Tenants.Maintenance/Views/Admin/Index.cshtml
wAsnk: I did something like that. Now auto-refresh will start once New version and Continue button is clicked.

BenedekFarkas: Nice!

BenedekFarkas (approved):

BenedekFarkas:
@wAsnk after some further thought/testing I figured we can simply the auto-refresh and run it as long as the toggle is on and don't care about anything else. What do you think? 7fc3bc0d45e9532656131f22c18f0b71de6d8e57

wAsnk:
> @wAsnk after some further thought/testing I figured we can simply the auto-refresh and run it as long as the toggle is on and don't care about anything else. What do you think? [7fc3bc0](https://github.com/Lombiq/Hosting-Tenants/commit/7fc3bc0d45e9532656131f22c18f0b71de6d8e57)

Yeah, that would be the simplest solution, so we don't have to do all that "workaround".

Lombiq.Hosting.Tenants.Maintenance/Views/StaggeredTenantsWakeUpAdmin/StaggeredTenantWakeUpDetails.cshtml
```diff
@@ -0,0 +1,165 @@
...
+    var part = ((ContentItem)Model.ContentItem).As<StaggeredTenantWakeUpPart>();
+    var currentPercentage = part.ProgressPercentage;
+    var started = part.Paused == false && part.StartedUtc != null && part.FinishedUtc == null;
+}
+
+<div class="buttons mt-4">
+    @await DisplayAsync(await New.StaggeredTenantWakeUp__NewVersionButton(Part: part))
+    @await DisplayAsync(await New.StaggeredTenantWakeUp__ContinueButton(Part: part))
+    @await DisplayAsync(await New.StaggeredTenantWakeUp__PauseButton(Part: part))
+    @await DisplayAsync(await New.StaggeredTenantWakeUp__EditButton(Part: part))
```
BenedekFarkas: @wAsnk Is there a particular reason for these being in separate shapes/templates?

wAsnk: It made sense to separate them, this view is already huge. No particular reason.

BenedekFarkas: Are you OK with me moving them to `StaggeredTenantWakeUpDetails.cshtml` instead? The statically typed parameters for the routing could be shared in a variable then. And we wouldn't need these as separate shapes (for overriding).
I don't think line count matters here that much.

wAsnk: Ofc I have no problem with that!

BenedekFarkas (approved):

PR #194, opened 2026-08-29
NEST-608: Elasticsearch rebuild workarounds

sarahelsaig:
[NEST-608](https://lombiq.atlassian.net/browse/NEST-608)

[NEST-608]: https://lombiq.atlassian.net/browse/NEST-608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.Hosting.Tenants.Maintenance.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
+++ b/Lombiq.Hosting.Tenants.Maintenance.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
@@ -126,6 +126,12 @@ public static class TestCaseUITestContextExtensions
                 TimeSpan.FromMinutes(2),
                 cancellationToken: context.Configuration.TestCancellationToken);
 
+            var maintenanceManager = serviceProvider.GetRequiredService<IMaintenanceManager>();
+            var latestMaintenanceExecution = await maintenanceManager.GetLatestExecutionByMaintenanceIdAsync(
+                ChangeUserSensitiveContentMaintenanceProvider.ProviderId);
+            latestMaintenanceExecution.ShouldNotBeNull();
+            latestMaintenanceExecution.IsSuccess.ShouldBeTrue(latestMaintenanceExecution.Error);
+
             var userManager = serviceProvider.GetRequiredService<UserManager<IUser>>();
 
             var testUser = (User)await userManager.FindByNameAsync(TestUser.UserName);
--- a/Lombiq.Hosting.Tenants.Maintenance.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
+++ b/Lombiq.Hosting.Tenants.Maintenance.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
@@ -4,6 +4,7 @@ using Lombiq.Hosting.Tenants.Maintenance.Maintenance.ChangeUserSensitiveContent;
 using Lombiq.Hosting.Tenants.Maintenance.Services;
 using Lombiq.Tests.UI.Constants;
 using Lombiq.Tests.UI.Extensions;
+using Lombiq.Tests.UI.Helpers;
 using Lombiq.Tests.UI.Models;
 using Lombiq.Tests.UI.Services;
 using Microsoft.AspNetCore.Identity;
@@ -120,12 +121,10 @@ public static class TestCaseUITestContextExtensions
             // Since the processing is not blocking the main thread, we have to wait until the queue is emptied in the
             // background.
             var queue = serviceProvider.GetRequiredService<IChangeUserSensitiveContentQueue>();
-            var waiting = true;
-            for (var i = 0; waiting && i < 120; i++)
-            {
-                waiting = queue.Count > 0;
-                await Task.Delay(1000, context.Configuration.TestCancellationToken);
-            }
+            ReliabilityHelper.DoWithRetriesOrFail(
+                () => queue.Count == 0,
+                TimeSpan.FromMinutes(2),
+                cancellationToken: context.Configuration.TestCancellationToken);
 
             var userManager = serviceProvider.GetRequiredService<UserManager<IUser>>();
 
--- a/Lombiq.Hosting.Tenants.Maintenance.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
+++ b/Lombiq.Hosting.Tenants.Maintenance.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
@@ -117,13 +117,24 @@ public static class TestCaseUITestContextExtensions
 
         await context.Application.UsingScopeServiceProviderAsync(async serviceProvider =>
         {
+            // Since the processing is not blocking the main thread, we have to wait until the queue is emptied in the
+            // background.
+            var queue = serviceProvider.GetRequiredService<IChangeUserSensitiveContentQueue>();
+            var waiting = true;
+            for (var i = 0; waiting && i < 120; i++)
+            {
+                waiting = queue.Count > 0;
+                await Task.Delay(1000, context.Configuration.TestCancellationToken);
+            }
+
             var userManager = serviceProvider.GetRequiredService<UserManager<IUser>>();
 
             var testUser = (User)await userManager.FindByNameAsync(TestUser.UserName);
-            testUser.UserName.ShouldBe(TestUser.UserName);
+            testUser!.UserName.ShouldBe(TestUser.UserName);
             testUser.Email.ShouldBe(TestUser.Email);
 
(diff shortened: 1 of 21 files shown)


Lombiq.Hosting.Tenants.Maintenance/Maintenance/ElasticsearchIndices/RebuildElasticsearchIndexesMaintenanceProvider.cs
```diff
@@ -12,14 +13,14 @@
 
 namespace Lombiq.Hosting.Tenants.Maintenance.Maintenance.ElasticsearchIndices;
 
-public class RebuildElasticsearchIndicesMaintenanceProvider : MaintenanceProviderBase
+public class RebuildElasticsearchIndexesMaintenanceProvider : MaintenanceProviderBase
```
sarahelsaig, line 13: ~Changing the name is necessary to unblock the new behavior.~ Also to use consistent US spelling. :) There are no downsides to this change

EDIT: upon further examination, I don't think it's strictly necessary but there is no harm and the spelling is still better so I suggest keeping it.

sarahelsaig:
lol breaking changes
There are actually [less](https://github.com/Lombiq/Hosting-Tenants/pull/194/changes#diff-b5a53e070ce0e6e6aea0466b1205e8a4f4f6aa7faea8b7a8e67d3cdd030d8050) breaking changes now.

Lombiq.Hosting.Tenants.Maintenance.Tests.UI/Extensions/TestCaseUITestContextExtensions.cs
```diff
@@ -117,13 +117,24 @@ public static async Task ChangeUserSensitiveContentMaintenanceExecutionAsync(thi
...
         {
+            // Since the processing is not blocking the main thread, we have to wait until the queue is emptied in the
+            // background.
+            var queue = serviceProvider.GetRequiredService<IChangeUserSensitiveContentQueue>();
+            var waiting = true;
+            for (var i = 0; waiting && i < 120; i++)
+            {
+                waiting = queue.Count > 0;
+                await Task.Delay(1000, context.Configuration.TestCancellationToken);
+            }
```
Piedone: Use `ReliabilityHelper` instead because it also supports a straightforward process timeout (what's more important for tests than iteration count) and throws if it elapsed.

Lombiq.Hosting.Tenants.Maintenance/Maintenance/ChangeUserSensitiveContent/ChangeUserSensitiveContentQueue.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using YesSql;
+
+namespace Lombiq.Hosting.Tenants.Maintenance.Maintenance.ChangeUserSensitiveContent;
+
+public class ChangeUserSensitiveContentQueue : IChangeUserSensitiveContentQueue
+{
+    private readonly IServiceScopeFactory _serviceScopeFactory;
+    private readonly ConcurrentQueue<long> _userIds = new();
```
Piedone, line 13: It's no issue that this is volatile and will vanish if the app is restarted mid-maintenance, right?

Lombiq.Hosting.Tenants.Maintenance/Maintenance/ChangeUserSensitiveContent/ChangeUserSensitiveContentBackgroundTask.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using YesSql;
+using static Lombiq.HelpfulLibraries.OrchardCore.Users.PasswordHelper;
+
+namespace Lombiq.Hosting.Tenants.Maintenance.Maintenance.ChangeUserSensitiveContent;
+
+[BackgroundTask(
+    Schedule = "* * * * *",
+    Description = "Check on the .")]
```
Piedone: Image

Lombiq.Hosting.Tenants.Maintenance/Maintenance/ChangeUserSensitiveContent/ChangeUserSensitiveContentBackgroundTask.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        // We don't want to login with these accounts, so we are generating the same password hash for each user, to
+        // make the process faster.
+        var passwordHash = passwordHasher.HashPassword(new User(), GenerateRandomPassword(32));
+
+        await ExecuteAsync(session, firstBatch, randomNameGenerator, passwordHash, cancellationToken);
+        while (await _queue.DequeueAsync() is { Count: > 0 } batch)
+        {
+            await ExecuteAsync(session, batch, randomNameGenerator, passwordHash, cancellationToken);
+        }
```
Piedone: This potentially long loops looks like an issue with a huge number of users, since then eventually this will time out. Shouldn't be a problem until hundreds of thousands, though.

Piedone (changes requested):

Lombiq.Hosting.Tenants.Maintenance/Maintenance/ChangeUserSensitiveContent/ChangeUserSensitiveContentQueue.cs
sarahelsaig: Good point. I've altered the maintenance provider to finish in a failed state. If the background task finishes, it updates the maintenance execution to be successful instead. Also updated the configuration to exclude the email domain used in the new values. This way even if the app or tenant is restarted mid-process, it will only try to work on the users that weren't updated.

EDIT: I've tested this in `https://dotnest-com-staging.lombiq.com/` and works as expected. Upon reset from prod, it launches the maintenance and it looks like this:
image
I was able to go into Admin > Access Control > Users and see it replace the users at lightning pace.
Once it was done, the background task updated the index too:
image

Lombiq.Hosting.Tenants.Maintenance/Maintenance/ChangeUserSensitiveContent/ChangeUserSensitiveContentBackgroundTask.cs
sarahelsaig: This background task is invoked by `OrchardCore.ModulesModularBackgroundService`, which is a type of `Microsoft.Extensions.Hosting.BackgroundService`. [According to the Microsoft docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.backgroundservice.executeasync?view=net-11.0-pp#microsoft-extensions-hosting-backgroundservice-executeasync(system-threading-cancellationtoken):~:text=The%20implementation%20should%20return%20a%20task%20that%20represents%20the%20lifetime%20of%20the%20long%20running%20operation%28s%29%20being%20performed) this is specifically made for long-running tasks. The execution passes in a `CancellationToken stoppingToken`, but that's [only triggered during application shutdown](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostedservice.stopasync?view=net-11.0-pp#microsoft-extensions-hosting-ihostedservice-stopasync(system-threading-cancellationtoken):~:text=Triggered%20when%20the%20application%20host%20is%20performing%20a%20graceful%20shutdown) so I don't think there is any timeout in this context.

Piedone: The loop is in a single DB context/transaction. `SaveAsync` flushes it after the batch though at least, right?

Piedone (approved):

Lombiq.Hosting.Tenants.Maintenance/Maintenance/ChangeUserSensitiveContent/ChangeUserSensitiveContentBackgroundTask.cs
sarahelsaig: I think SaveAsync flushes, so each loop is a separate transaction. I didn't write that part, but I figure that's why we were updating in 15 user batches in the fist place? The maintenance was originally running in the main thread and that didn't cause a timeout with over 8000 users, so we will be fine for sure now that it's running in the background. Once we have our own Facebook-conquering social media we will have to stop using this maintenance. :)

Piedone: Yeah, as I said this shouldn't be a problem in any case for now.

